PR #1, opened 2018-08-14
[WIP] [rpc] Wrap grpc server for grpc-web compatability

graup:
This uses a [grpc-web wrapper](https://github.com/improbable-eng/grpc-web#why) to support browsers.

For review by @kjunu as I rewrote some of his code.

grpc-web is a protocol extention over plain grpc. See https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md

Maybe we need to keep support for the "basic" GRPC protocol for the Java client? Maybe publish them on different ports? I guess @bylee needs to test this.

Companion PR: https://github.com/aergoio/herajs/pull/2

### DIFF
--- a/config/types.go
+++ b/config/types.go
@@ -36,8 +36,8 @@ type BaseConfig struct {
 // RPCConfig defines configurations for rpc service
 type RPCConfig struct {
 	// RPC and REST
-	NetServiceAddr string `mapstructure:"netserviceaddr" description:"RPC N2C network service address"`
-	NetServicePort int    `mapstructure:"netserviceport" description:"RPC N2C network service port"`
+	NetServiceAddr string `mapstructure:"netserviceaddr" description:"RPC service address"`
+	NetServicePort int    `mapstructure:"netserviceport" description:"RPC service port"`
 	// RPC API with TLS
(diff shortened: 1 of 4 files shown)


graup:
I refactored the code to serve two different servers on different ports (keeping the original port for the original implementation). Please someone review as I am not very familiar with the naming conventions.

hayarobi:
@ashen1dev @kjunu @graup 
There was an informal agreement that aergo server will serve grpc and REST API with same port. When I surveyed https://github.com/grpc-ecosystem/grpc-gateway at that time, it seemed possible to do by checking header. 
We need to discuss this in the meeting.

graup:
@hayarobi I found a solution using cmux (multiplexer on TCP level). I tested both connection types with the JS client.

By the way, there is a big performance difference.

Test: sign and send 500 transactions.

Over grpc (HTTP2):
Transaction sign: 372.702259ms
Transaction commit: 85.922567ms

Over grpc-web (HTTP 1):
Transaction sign: 945.394124ms
Transaction commit: 446.806474ms

It's good to support both so clients can choose the faster version if possible.

graup:
Now we'll need to add the REST api to the same HTTP server that currently handles grpc-web.

PR #2, opened 2018-08-20
Topic/chain

everjs78:
please review for new reorg logic


state/statedb.go
```diff
@@ -209,8 +210,9 @@ func (sdb *ChainStateDB) Rollback(blockNo types.BlockNo) error {
 	sdb.Lock()
 	defer sdb.Unlock()
 
-	for sdb.latest.BlockNo > blockNo {
-		bs, err := sdb.loadBlockState(sdb.latest.BlockHash)
+	target := sdb.latest
```
qcollapse: The variable name `target` looks confusing since there are two targets: the last rollback target and the final target.

PR #3, opened 2018-08-21
[RPC] Matching http2 with writer for java-grpc compatibility

acktsap:
I found that java gRPC client is not compatible with current matcher.

According to [cmux repository](https://github.com/soheilhy/cmux), use MatchWithWriter when using grpc java client.

But we have to check MatchWithWriter version is compatible with [herajs](https://github.com/aergoio/herajs). Maybe @graup have to test it.


graup:
I tested it, it works with both the Node.js and Web version of herajs.

PR #13, opened 2018-11-16
WIP: ES Indexer plugin

graup:
This is an example plugin with some necessary changes in the core.

This PR is just for review, the plugin will live in its own repository once we have a plugin infrastructure.

In particular I wonder if adding a RequestBroadcast method like I did makes sense.


graup:
After some discussion it's clear that this is not a good way to do what I wanted. Instead of a broadcast, there should be some sort of publish/subscribe model.

ashen1dev:
As we discussed, I'll close this PR. Thanks.

Issue #16, opened 2018-11-28
Decimal Support: 9 or 18


ashen1dev:
'AERGO' is main coin of Aergo Network. One advantage of cryptocurrency is that it can be divided smaller than the base unit.(Better Liquidity) and we want to define the minimum unit called AER. Aergo's smart contracts, SDK and AergoCLI will  use AER as a basic unit. The problem is how much AER is smaller than AERGO. In initial discussion, we noticed that most Ether wallets actually use the GWei unit as a fee unit.

So we tried to define AER to be 10 ^ -9 of AERGO (until Release 0.8.1). But soon, we found the compatibility problem. In case of the ERC20 token transfer from the Ethereum to AERGO main network(via the Asset Bridge), token value will be truncated from 18 decimals to 9 decimals.

We will define AER as 10 ^ -18. However, in this case, we need to implement a separate big number data structure instead of using unsigned int 64 bits.

In addition, it is also better to be compatible with Ethereum by supporting up to 10 ^ -18 in terms of 'Forward Compatibility'.

graup:
For compatibility reasons, I think it makes sense to stick with 18 digits. There are even some proposals to make that compulsory for all ERC20 tokens (https://github.com/ethereum/EIPs/issues/724), but anyway it is the de-facto standard.

We'll need to refactor all code regarding balances and transaction amounts in the server and all clients.

graup:
[x] Herajs merged support for 18 decimals in https://github.com/aergoio/herajs/commit/7435563b9679b4816010beb03ad89e7b14f87b71

ashen1dev:
We decided to go on 18 decimals.

PR #18, opened 2018-12-04
Change uint64 to bytes for supporting big number

kjunu:
https://github.com/aergoio/aergo/issues/16


account/accountservice_test.go
```diff
@@ -155,7 +156,7 @@ func TestVerfiyFail(t *testing.T) {
 	assert.NotNil(t, tx.Body.Sign, "failed to sign")
 
 	//edit tx after sign
-	tx.Body.Amount = 0xff
+	tx.Body.Amount = new(big.Int).SetUint64(0xff).Bytes()
```
graup: Isn't `big.NewInt(0xff).Bytes()` the same as `new(big.Int).SetUint64(0xff).Bytes()`?

cmd/aergocli/cmd/staking.go
```diff
@@ -16,55 +18,56 @@ import (
...
+		return errors.New("Failed: minimum staking value is " + types.StakingMinimum.String())
 	}
 
 	tx := &types.Tx{
 		Body: &types.TxBody{
 			Account:   account,
 			Recipient: []byte(types.AergoSystem),
-			Amount:    amount,
-			Price:     0,
+			Amount:    amountBigInt.Bytes(),
```
graup: Just a question, what is the point of calling Bytes()? Is it to create a copy?

account/accountservice_test.go
kjunu: right 👍 I will fix :)

cmd/aergocli/cmd/staking.go
kjunu: yes it's copy :(

https://github.com/golang/go/blob/master/src/math/big/int.go#L440

PR #22, opened 2018-12-14
[cli] keygen --json option to export address and encrypted privkey

graup:
There's currently no easy way to just generate a private key and address offline (without a server and internet connection). Herajs and Herajs can do that, of course, but if someone just want to quickly make a new account, that's not that easy to use.

I figured that `aergocli keygen` does pretty much exactly that, but it didn't export the new key in the format used by other tools. So I added the option `--json --pasword test` so that the new key's address and encrypted key are exported.

### DIFF
--- a/account/key/store.go
+++ b/account/key/store.go
@@ -79,6 +79,11 @@ func (ks *Store) ExportKey(addr Address, pass string) ([]byte, error) {
 	if key == nil {
 		return nil, err
 	}
+	return EncryptKey(key, pass)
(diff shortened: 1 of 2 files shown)


graup:
Question: maybe it would be better to ONLY generate the json file when `--json` is passed in? I guess having raw private keys lying around could be a security issue if people use this tool who are not that familiar with these things. The json file's privkey is encrypted, so slightly better in terms of security.

PR #24, opened 2018-12-14
Named account

kjunu:
Support named account
* Name - address mappings are stored in "aergo.name" contract state DB


graup:
At first glance, looks good to me.

This introduces new protocols, do we have these documented somewhere? We should probably have a common specification format for the governance tx types. Esp. since `c12134567,23456890` is kind of arbitrary.

Maybe a good idea to add a separate page to the new dev portal / documentation with format specifications?

kjunu:
https://github.com/aergoio/aergo/wiki/transaction-types

PR #61, opened 2019-03-08
[aergosvr] Changes on aergosvr options

graup:
- behaviors of running --testmode
     * on empty chain directoroy
         A chain with "dev.chain" as Magic of ChainID will be generated
     * on some chain directory
         only works if chainID's Magic is "dev.chain", otherwise exits

  - changes aergosvr init command
     * remove --dir option for init command, use --home instead
     * also create default config.toml in home directory

  - Minor refactoring
     * spliting init and version into seperate files
     * enabletestmode, usetestnet option not to be written in config file


graup:
Targeted for release 0.12.1

PR #68, opened 2019-04-16
[aergo.name] Bugfix, add missing error handling

kjunu:
This code need to reset or rollback testnet


graup:
looks ok to me. thanks for adding the test.

PR #69, opened 2019-04-22
Topic/crashrecover with hotfix

everjs78:
Please, don't merge this pr before review by Mr.Woo

implemented crash recovery 
[https://blocko.atlassian.net/wiki/spaces/~456868988/pages/623150290/Crash+Recovery+Error+block
](url)
1. normal case
- nothing to do, because state db is changed but it doesn't need recovery.
2. reorg case 
-Situation with reorg marker remaining
 After the rollforward, re-execute the operations to modify the chain db.

### DIFF
--- a/chain/reorg.go
+++ b/chain/reorg.go
@@ -10,7 +10,6 @@ import (
 	"github.com/aergoio/aergo/internal/enc"
 	"github.com/aergoio/aergo/message"
 	"github.com/aergoio/aergo/types"
-	"os"
 )
 
 const (
@@ -140,9 +139,7 @@ func (cs *ChainService) reorg(topBlock *types.Block, marker *ReorgMarker) error 
 	if err := reorg.swapChain(); err != nil {
 		switch ec := err.(type) {
 		case *ErrDebug:
-			if len(os.Getenv("DEBUG_CHAIN_CRASH")) == 0 {
-				return ec
-			}
+			return ec
 		}
 		logger.Fatal().Err(err).Msg("reorg failed while swapping chain, it can't recover")
 		return err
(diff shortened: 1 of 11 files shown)


chain/reorg.go
```diff
@@ -283,44 +414,83 @@ func (reorg *reorganizer) gatherChainInfo() error {
...
+	cdb := reorg.cs.cdb
+
+	startBlock := reorg.brStartBlock
+	bestBlock := reorg.bestBlock
+	topBlock := reorg.brTopBlock
+
+	reorg.brStartBlock = startBlock
+	reorg.bestBlock = bestBlock
+
+	for tmpBlk := bestBlock; tmpBlk.GetHeader().GetBlockNo() > startBlock.GetHeader().GetBlockNo(); {
```
qcollapse: The two loops in this method are almost the same. The duplicates would rather be removed.

everjs78: I refactored that part.

chain/reorg.go
```diff
@@ -125,13 +193,65 @@ func (cs *ChainService) reorg(topBlock *types.Block) error {
 //  - tx mapping
 //  - best block
 func (reorg *reorganizer) swapChain() error {
-	cs := reorg.cs
+	cdb := reorg.cs.cdb
 
 	logger.Info().Msg("swap chain to new branch")
 
-	cs.cdb.swapChain(reorg.newBlocks)
+	marker := NewReorgMarker(reorg)
```
qcollapse: Is a new marker needed for a recovery case?

In the case of a recovery, the marker seems to be loaded from the chain DB and dumped. In my view, here we check whether reorg has a valid marker in it and (if it has) use it instead of creating a new one.

everjs78: I have refactored reorg marker as you have advised. 
I added read/write api to marker. and I fixed that marker is made in gatherPostFn after gather() during reorg.
Thanks!

PR #70, opened 2019-04-24
[CHAIN] fix reorg bug & refactoring

everjs78:
1. fix bug where old tx mapping could not committed when swapping tx mapping.
    old txs has saved in transaction but not commited if tx count < 10000.
    but when rpc is called, tx mapping info is dropped if block of tx isn't included in chain 
2. refactoring - apply bulk transaction when swapping chain info during reorg


qcollapse:
It looks ok! Let's merge it.

PR #75, opened 2019-05-14
[rpc] Add blocksize to BlockMetadata (Fixes #73)

### LINKED ISSUE #73: Add block size to API
Please add a field `size` to both `BlockMetadata` and `Block`, included in the return types of `ListBlockMetadata`, `ListBlockMetadataStream`, `GetBlockMetadata`, `GetBlock`, `ListBlockStream`.

The field should contain the value returned by [`block.Size()`](https://github.com/aergoio/aergo/blob/b56738252fd8ebfda47f41fe28d30d19c65126c4/types/blockchain.go#L334) or similar.

Question: The referenced method Size() seems to loop over all transactions, I wonder if that is very efficient. Maybe there is a more efficient way?

graup:
Also refactored the generation of the BlockMetaData object into a function in the Block interface.

Goes along with this protobuf change: https://github.com/aergoio/aergo-protobuf/pull/26

Fixes #73


graup:
Process:

1. Review and merge https://github.com/aergoio/aergo-protobuf/pull/26
2. Review this PR. Don't merge.
3. Update this PR with the correct reference to aergo-protobuf
4. Merge this PR

hayarobi:
There is a method named Block.Size(). Please see around line 330 in types/blockchain.go. 

The Size() method seems not to returning exact size, but please consider to use it, or not.

graup:
@hayarobi in #73, @qcollapse suggested to use `proto.Size`. I'm not sure which one would be better. What do you think which value would be more useful for admins or people analyzing blockchain data?

graup:
After discussion, I think `proto.Size(block)` is the easiest. This is also easily reproducible in the client. I only add it to BlockMetadata. For Block, clients can calculate it on their own.

PR #78, opened 2019-05-16
[RPC] Support to get name at a specific number of block

kjunu:
relate : https://github.com/aergoio/aergo-protobuf/pull/27


graup:
looks ok to me

graup:
I especially like the use of the `AccountStateReader` interface :+1:

PR #83, opened 2019-06-12
[ctr] Add the redeploy transaction

mlogue76:
- contracts can only be updated on a private network


contract/contract.go
```diff
@@ -151,14 +152,25 @@ func preLoadWorker() {
...
 		recipient := txBody.Recipient
 
 		if txBody.Type != types.TxType_NORMAL || len(recipient) == 0 {
 			continue
 		}
 
+		if reqInfo.current.GetBody().Type == types.TxType_REDEPLOY {
+			currentTxBody := reqInfo.current.GetBody()
+			if bs != nil {
+				delete(bs.CodeMap, types.ToAccountID(currentTxBody.Recipient))
```
eve2adam: need delete when next is redeploy

PR #94, opened 2019-11-11
Reduce log level for some messages in chain and rpc

graup:
Reduced log level for some messages that seem unnecessary in normal operation.


graup:
Ping @kjunu @everjs78

kjunu:
ok :)

PR #102, opened 2020-01-22
Topic/keystore

acktsap:
Since store module is too complicated, i've just refactored to extract storage interface.
This is a pre-stage of keystore. @graup have to review it to check side effects of refactoring.

After keystore is implemented. It will be merged. Don't merge until keystore is implemented.


acktsap:
And i also add keystore proposal

# Proposal

We have no common keystore file format among clients. So i propose a common keystore format. A foramt is inspired by ethereum keystore file.

## File

AmLxM6Ae1fUQKQ9AJbT2sv54Mzv96jJM4pKycCzpuAjNj4564eFs__keystore.txt

```json
{
  "aergo_address": "Amdxxxx",
  "ks_version": 1,
  "kdf": {
    "algorithm": "algorithm_name",
    "params": {
      "some_param": "some_value"
    },
    "mac": "mesasge_authentication_code"
  },
  "cipher": {
    "algorithm": "algorithm_name",
    "params": {
      "some_param": "some_value"
    },
    "ciphertext": "dxxxx",
  }
}
```

### Rule

- Version field is for choosing cipher algorithm. If version updates, cipher algorithm would be updated (also format).
- Every file has it's own algorithm and parameters per version. If any of the parameters break, throw error (Invalid file format).
- File name is '{account_address}__keystore.txt' or '{alias}__keystore.txt'
- File name is used to identify keystore file. No same address (or alias) is allowed.
- For the alias file, you can have same keystore file representing same account using different alias. That's fine.
- Alias must be form of `[a-zA-Z0-9]+`

## KeyStore

Directory based keystore

```sh
> ls ./my_keystore
AmLxM6Ae1fUQKQ9AJbT2sv54Mzv96jJM4pKycCzpuAjNj4564eFs__keystore.txt
AmgVbUZiReUVFXdYb4UVMru4ZqyicSsFPqumRx8LfwMKLFk66SNw__keystore.txt
AmLxM6Ae1fUQKQ9AJbT2sv54Mzv96jJM4pKycCzpuAjNj4564eFs__keystore.txt
myalias1__keystore.txt
myalias2__keystore.txt
```

### Rule

- Identity is account_address or alias. Only file with format `{account_address or alias}__keystore.txt` is accepted
- Save : (identity, password)\
  If (identity is exists) -> throw error\
  else -> Encrypt with a password -> Save as file name `{account_address or alias}__keystore.txt`
- Load : (identity, password)\
  If (Identify by file name fails) -> throw error\
  else If (decrypt with a password failure) -> throw error\
  else -> decrypt result
- List : List existing identities.

## V1 encrypt/decrypt process

Encrypt

```go
hashedPassword := Script(rawPassword)
cipherText := encrypt(rawPrivateKey, hashedPassword)
mac := calculateMac(cipherText, hashedPassword)
```

Decrypt

```go
var json // json file

if json.ks_version != 1 {
  return nil, "Invalid keystore version"
}

hashedPassword := Script (rawPassword)
mac := (cipherText, hashedPassword)
if json.mac != mac {
  return nil, "Invalid mac"
}

rawPrivateKey := decrypt(cipherText, hashedPassword)
```

account/key/storage.go
```diff
@@ -0,0 +1,15 @@
+package key
+
+import (
+	"github.com/btcsuite/btcd/btcec"
+)
+
+type Identity = []byte
```
graup: Any reason for adding this new type instead of reusing `Address` (from key/address.go)?

account/key/utils.go
```diff
@@ -0,0 +1,60 @@
...
+	h.Write(b2)
+	return h.Sum(nil)
+}
+
+func encrypt(base, key, data []byte) ([]byte, error) {
+	block, err := aes.NewCipher(key)
+	if err != nil {
+		return nil, err
+	}
+	// Never use more than 2^32 random nonces with a given key because of the risk of a repeat.
```
graup: You added this comment, but I don't understand what it has to do with the following line.

cmd/aergocli/cmd/accounts.go
```diff
@@ -93,7 +93,7 @@ var newCmd = &cobra.Command{
 				cmd.Printf("Failed: %s\n", err.Error())
 				return
 			}
-			err = ks.SaveAddress(addr)
+			// err = ks.SaveAddress(addr)
```
graup: I think you're right, `CreateKey` calls `addKey` which calls `save`. Let's remove this whole line instead of commenting it.

account/key/storage.go
acktsap: I've just replaced Address with Identity in 1ce93068f62e294417aa5ab2e3778ae47ecc39d0

Address and Identity are both just a byte array. But the Address has less meaning. I intend to cover more meaning (eg. alias).

account/key/utils.go
acktsap: I didn't add it. I've just copied.

cmd/aergocli/cmd/accounts.go
acktsap: Yes you are right. I missed. Fixed in 1ce93068f62e294417aa5ab2e3778ae47ecc39d0

account/key/storage.go
graup: Okay, good!

account/key/utils.go
graup: ok maybe we can improve it later then

acktsap:
Build failure is fixed in another commit. @graup can you review it? Local build would not fail.

account/key/badgerdb.go
```diff
@@ -83,6 +114,17 @@ func (ks *BadgerStorage) Close() {
 	ks.db.Close()
 }
 
+func openBadgerDb(path string) (opened db.DB, err error) {
+	defer func() {
+		failed := recover()
```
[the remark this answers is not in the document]
acktsap: `db.NewDB(db.LevelImpl, path)` panic on error. I indend to provide similar interface (*OpenedDB, err) for all storage type.

account/key/aergo_storage.go
```diff
@@ -0,0 +1,147 @@
...
+	ks.RWMutex.RLock()
+	defer ks.RWMutex.RUnlock()
+
+	// FIXME: save itself.... need to refactor store.go
+	encodedIdentity := types.EncodeAddress(identity)
+
+	fileName := fmt.Sprintf(fileNameTemplate, string(encodedIdentity))
+	absFilePath := filepath.Join(ks.storePath, fileName)
+
+	// origin badger storage permits duplication...
```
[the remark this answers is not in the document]
acktsap: It makes absolute file path for corresponding identity (address) and check existance of file by path.

types/errors.go
```diff
@@ -54,6 +54,8 @@ var (
 
 	ErrShouldUnlockAccount = errors.New("should unlock account first")
 
+	ErrWrongIdentityPassWord = errors.New("identity or password is incorrect")
```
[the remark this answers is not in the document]
acktsap: Removed in ba6baba3bd0593673185b22608bbcd8d2fa90305

account/key/store.go
```diff
@@ -135,6 +149,15 @@ func (ks *Store) GetAddresses() ([]Identity, error) {
 	return ks.storage.List()
 }
 
+// can open & has at least on account in it
+func hasAccount(storage Storage) bool {
```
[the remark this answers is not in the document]
acktsap: Removed in ba6baba3bd0593673185b22608bbcd8d2fa90305

account/key/address.go
```diff
@@ -1,29 +1,19 @@
...
-	keyLength := len(pubkey.X.Bytes())
-	if keyLength < 32 { // Add padding
-		for i := 1; i < types.AddressLength-keyLength; i++ {
-			binary.Write(addr, binary.LittleEndian, uint8(0))
-		}
-	}
-	binary.Write(addr, binary.LittleEndian, pubkey.X.Bytes())
-	return addr.Bytes() // 33 bytes
+	return delegate.GenerateAddress(pubkey)
 }
```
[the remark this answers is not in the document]
acktsap: Replaced in ba6baba3bd0593673185b22608bbcd8d2fa90305

account/key/crypto/test/AmMzdvQMremMBwFqrVWc4Urq1VaFPmbAGY7VBtfDnCXE93vSct7H__keystore.txt
```diff
@@ -0,0 +1,22 @@
+{
+  "aergo_address":"AmMzdvQMremMBwFqrVWc4Urq1VaFPmbAGY7VBtfDnCXE93vSct7H",
+  "ks_version":1,
```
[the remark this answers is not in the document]
acktsap: It looks "ks_version": "1" is right. Fixed in ba6baba3bd0593673185b22608bbcd8d2fa90305

graup:
Looks good to me so far. Will need some more QA testing after it's merged.

To summarize what happens currently: 
- If an old badger db exists, we keep using that one.
- Otherwise, creating/importing addresses now uses the new keystore (creating keystore files in the filesystem)

Should we provide some kind of migration path for users of the old system? For migration, you'd have to export all your old accounts (using each one's password), delete the db, and then re-import all accounts, correct? Maybe just providing a guide will be fine.

account/key/store.go
```diff
@@ -28,27 +27,39 @@ type Store struct {
...
 		unlocked:     map[string]*keyPair{},
 		unlockedLock: &sync.Mutex{},
-		storage:      db.NewDB(db.LevelImpl, dbPath),
 	}
+
+	// FIXME: more elegant coding
+	if storage, err := LoadBadgerStorage(storePath); nil == err {
+		store.storage = storage
+	} else {
+		storage, err := NewAergoStorage(storePath)
```
graup: This stores the keystore files directly in the `$datadir`, right? Maybe we should add a sub directory like `keystore`?

acktsap: Applied in 596508c5f479e6f05f37c2da8f334796826699b8

acktsap: Applied in 9e4226a9998191014d53cae1bceba1ab725f10d3

Issue #108, opened 2020-02-08
libgmp.so can get built into libtool/lib64


aspiers:
`contract/vm.go` seems to assume that libgmp.so will always get built into `libtool/lib/`:

https://github.com/aergoio/aergo/blob/87a379a44bb03c8dc6f29e70732e5a1b297a9a1f/contract/vm.go#L13

but on my (openSUSE) system, it gets built in `libtool/lib64`:

```
libtool/lib:
total 1.5M
-rw-r--r-- 1 adam users 471K Feb  8 23:19 liblmdb.a
-rwxr-xr-x 1 adam users 306K Feb  8 23:19 liblmdb.so*
-rw-r--r-- 1 adam users 741K Feb  8 23:19 libluajit-5.1.a
drwxr-xr-x 3 adam users 4.0K Feb  3 19:45 lua/
drwxr-xr-x 2 adam users 4.0K Feb  8 23:19 pkgconfig/

libtool/lib64:
total 528K
-rwxr-xr-x 1 adam users  932 Feb  8 23:19 libgmp.la*
lrwxrwxrwx 1 adam users   16 Feb  8 23:19 libgmp.so -> libgmp.so.10.3.2*
lrwxrwxrwx 1 adam users   16 Feb  8 23:19 libgmp.so.10 -> libgmp.so.10.3.2*
-rwxr-xr-x 1 adam users 521K Feb  8 23:19 libgmp.so.10.3.2*
```

This results in a build error:

```
# github.com/aergoio/aergo/contract
gcc: error: contract/../libtool/lib/libgmp.so: No such file or directory
make[3]: *** [CMakeFiles/aergosvr.dir/build.make:57: CMakeFiles/aergosvr] Error 2
make[2]: *** [CMakeFiles/Makefile2:260: CMakeFiles/aergosvr.dir/all] Error 2
make[1]: *** [Makefile:84: all] Error 2
make: *** [Makefile:32: all] Error 2
```

Changing the above line in `contract/vm.go` to:

```Make
#cgo !darwin,!windows LDFLAGS: ${SRCDIR}/../libtool/lib/libluajit-5.1.a ${SRCDIR}/../libtool/lib64/libgmp.so -lm
```

[...]

PR #110, opened 2020-02-09
Fix build to expect libgmp in libtool/lib64 (#108)

aspiers:
This fixes the build on my Linux machine, but it may break it on others, so someone should review carefully before merging.  As I mentioned in #108, there must be a better approach than this which doesn't require hard-coding this path segment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ Gopkg.lock
 # python cache
 venv
 __pycache__
+/libtool/lib64/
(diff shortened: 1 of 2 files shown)


aspiers:
OK so Travis shows that it breaks on Ubuntu Xenial :-(  So a smarter fix is needed that dynamically determines the libdir.

eve2adam:
committed at 79484c20b752081a6e41ef0957551b57174cbe54 by other way

PR #111, opened 2020-02-09
Make brick use private dummy chain to provide db module (#107)

### LINKED ISSUE #107: brick doesn't provide db module
[The brick README](https://github.com/aergoio/aergo/blob/develop/cmd/brick/README.md) says:

> * supports state db, sql

but if I try to test a contract which references `db`, I get:

    ERR execution fail error="[string \"0cca71fcf76698693693c6edc703ea42c82494bfafa1b6a14100aabaa7a1afb08f\"]:0: attempt to index global 'db' (a nil value)" cmd=deploy module=brick

I looked into the source, and I found:

https://github.com/aergoio/aergo/blob/87a379a44bb03c8dc6f29e70732e5a1b297a9a1f/cmd/brick/context/context.go#L30

which calls `OnPubnet` which sets `PubNet` to `true`:

https://github.com/aergoio/aergo/blob/87a379a44bb03c8dc6f29e70732e5a1b297a9a1f/contract/vm_dummy.go#L677-L684

This eventually prevents the `db` module from being loaded into the Lua environment:

https://github.com/aergoio/aergo/blob/87a379a44bb03c8dc6f29e70732e5a1b297a9a1f/contract/vm.c#L37-L39

I proved this theory by changing

```go
	chain, err := contract.LoadDummyChain(contract.OnPubNet)
```

to:

```go
	chain, err := contract.LoadDummyChain()
```

and recompiling, and then the error goes away.

As a side note, this suggests that brick is missing automated test coverage to ensure that it is correctly providing the built-in contract modules.  I might file a separate issue for that.

aspiers:
The `db` module is only supported on non-public networks, so don't configure the dummy chain as a public network.  This may have other negative side-effects I'm not aware of, so please review carefully before merging.

Fixes #107.

### DIFF
--- a/cmd/brick/context/context.go
+++ b/cmd/brick/context/context.go
@@ -27,7 +27,7 @@ type context struct {
 }
 
 func Reset() {
-	chain, err := contract.LoadDummyChain(contract.OnPubNet)
+	chain, err := contract.LoadDummyChain()
 	if err != nil {
 		panic(err)
 	}


sunpuyo:
Currently brick is built on the public chain and does not support the db module. 
In the private chain, there is no tx fee, so the operation is different.
So it is necessary to support separate build or switch command to private mode.

aspiers:
[**@sunpuyo**](https://github.com/sunpuyo) commented on [February 10, 2020 7:07 AM](https://github.com/aergoio/aergo/pull/111#issuecomment-583984709 "2020-02-10T07:07:27Z - Replied by Github Reply Comments"):
> Currently brick is built on the public chain and does not support the db module. 
> In the private chain, there is no tx fee, so the operation is different.
> So it is necessary to support separate build or switch command to private mode.

How about introducing a new `-p` option to enable private mode?

aspiers:
[**@aspiers**](https://github.com/aspiers) commented on [February 10, 2020 11:10 AM](https://github.com/aergoio/aergo/pull/111#issuecomment-584071778 "2020-02-10T11:10:10Z - Replied by Github Reply Comments"):
> How about introducing a new `-p` option to enable private mode?

Any thoughts on this?  It would be extremely useful for us ... ;-)

mlogue76:
I will consider supporting it.

aspiers:
Thanks!

mlogue76:
Fixed in 175ee65b (develop branch)

aspiers:
Great, thank you so much! :grin:

PR #112, opened 2020-02-09
Ensure brick has a non-zero exit code on error

aspiers:
Always return non-zero exit code on error for consistency.  This ensures that if brick is called from within shell scripts / Makefiles etc., errors can be properly detected and handled.


graup:
@sunpuyo do you want to merge this for 2.2? We're going to make the release branch soon.

sunpuyo:
@graup Yes, this issue is completed and I merge this.

PR #114, opened 2020-02-10
[cli,rpc] add account import and export for keystore

graup:
- `account export` (breaking change) now returns keystore format by default
- `account export --wif` (new) returns legacy format
- `account import --keystore` (new) accepts keystore format for import
- `account import --if` (unchanged)

Also added the corresponding GRPC endpoints (https://github.com/aergoio/aergo-protobuf/pull/36):

- `ImportAccount`: `ImportFormat` has a new parameter `Keystore`. The endpoint selects the method based on existence of either the Wif or the Keystore parameter in the protobuffer msg.
- `ExportAccountKeystore` (new)
- `ExportAccount` (unchanged)


account/key/aergo_storage.go
```diff
@@ -88,29 +86,47 @@ func (ks *AergoStorage) Save(identity Identity, passphrase string, key *PrivateK
...
 	if nil != err {
 		return nil, types.ErrWrongAddressOrPassWord
 	}
-
-	// TODO: dispatch per keystore version
-	strategy := version2Strategy[encryptVersion]
-	privateKey, err := strategy.Decrypt(encrypted, passphrase)
+	privateKey, err := LoadKeystore(encrypted, passphrase)
 	if nil != err {
 		return nil, types.ErrWrongAddressOrPassWord
```
acktsap: Looks should return err.

acktsap:
Looks fine. But need to fix build failure

account/key/aergo_storage.go
graup: Agreed. Fixed 
2a95dc9

PR #118, opened 2020-02-13
support keystore in aergocli

eve2adam:
This allows local keystore to be used with all account commands (incl sendtx etc).

BREAKING CHANGE
- The new default behavior is to use a local keystore at `$HOME/.aergo`
- You can change the keystore directory by supplying `--keystore /path/to/keystore/`
- To restore the old behavior (using a server provided keystore), pass `--node-keystore`

Remote keystore functionality will be removed in future versions. The reason is that this is prone to unsafe misuse - if used with a publicly accessible node, this can expose user information to the public.

### DIFF
--- a/cmd/aergocli/cmd/root.go
+++ b/cmd/aergocli/cmd/root.go
@@ -54,10 +54,10 @@ var (
 
 	staking bool
 
-	remote           bool
-	importFormat     string
-	keystoreFilePath string
-	exportAsWif      bool
+	remote         bool
+	importFormat   string
+	importFilePath string
+	exportAsWif    bool
 
 	rootConfig CliConfig
 
@@ -82,6 +82,7 @@ func init() {
 	rootCmd.PersistentFlags().StringVar(&keyFile, "tlskey", "", "client key file for TLS ")
 	rootCmd.PersistentFlags().StringVarP(&host, "host", "H", "localhost", "Host address to aergo server")
 	rootCmd.PersistentFlags().Int32VarP(&port, "port", "p", 7845, "Port number to aergo server")
+	rootCmd.PersistentFlags().StringVar(&dataDir, "keystore", "$HOME/.aergo", "Path to keystore")
 }
 
 func initConfig() {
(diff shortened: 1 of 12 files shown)


cmd/aergocli/cmd/root.go
```diff
@@ -82,6 +82,7 @@ func init() {
 	rootCmd.PersistentFlags().StringVar(&keyFile, "tlskey", "", "client key file for TLS ")
 	rootCmd.PersistentFlags().StringVarP(&host, "host", "H", "localhost", "Host address to aergo server")
 	rootCmd.PersistentFlags().Int32VarP(&port, "port", "p", 7845, "Port number to aergo server")
+	rootCmd.PersistentFlags().StringVar(&dataDir, "keystore", "$HOME/.aergo", "Path to keystore")
```
graup: I wonder if it would be better if the default would be `""` for now, otherwise this immediately breaks all current uses. We can change the default in future versions (with previous announcement).

graup:
This doesn't seem to work. If you run e.g. `./bin/aergocli account stake --address AmNKKMAPiKr6mgWMRWXvajzEquNjsLPjeAvU7EpNcN9m4vunPmZU --amount 1000 --keystore ./wallet2/` I get this error:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x47c3f14]

goroutine 1 [running]:
github.com/aergoio/aergo/cmd/aergocli/cmd.fillChainId(0xc000276680, 0x0, 0xc000193450)
	/Users/paulgrau/projects/aergo/aergo/cmd/aergocli/cmd/contract.go:370 +0x44
```

I suspect it is because the `client` is not intiialized when `--keystore` is passed (cmd/aergocli/cmd/accounts.go:451)

graup:
There's also a bug if you call `account list` with an empty keystore (which is now the default). The fix for that would be to change this line https://github.com/aergoio/aergo/blob/623afd534b5004ff8551b079f287fe562a9cf1ce/cmd/aergocli/cmd/accounts.go#L136

to

    } else if addrs != nil && len(addrs) > 0 {

cmd/aergocli/cmd/root.go
eve2adam: it decided to use the default local keystore from v2.2 at team metting. so at this version release we need to announce the default change and how use node keystore like previous version

graup: Alright. It's usually better to announce breaking changes in advance, especially with minor releases.
So how do you use it with the old behavior? `--keystore ""`? Seems a bit counter-intuitive to have to provide an empty string. How about `--remote-keystore` as a flag which sets keystore to ""?

eve2adam: added node-keystore flag

aspiers:
AFAICS this has broken `account export`:

```
aergocli -H $mynode account export --node-keystore --address $addr --password foobar
rpc error: code = Unimplemented desc = unknown method ExportAccountKeystore for service types.AergoRPCService
```

graup:
This error appears when you're connecting to a server that's not upgraded
yet (for keystore export its using a new RPC method).

aspiers:
Ahhh that makes sense, thanks!

PR #128, opened 2020-04-08
[WIP] Feature/aergocli delete account

hanlsin:
New feature to delete an account when to export the account.

Without this, even if I exported the account and import in another device or wallet, I should take care the account at the machine.


hanlsin:
> need to fix testcase, error ocurred

Yup. I found that I missed one more part to apply.

PR #154, opened 2022-03-19
add bignum.ispositive()

kroggen:


kroggen:
This could be renamed to `bignum.ispositive()`

And then create a clone of the `bignum.isneg()` and name it `bignum.isnegative()` (both will exist, but the docs would mention only the last one)

kroggen:
@hayarobi Can I add the changes above on this PR?

hayarobi (approved):
lgtm

PR #165, opened 2022-05-17
Use the contract address on error messages

### LINKED ISSUE #113: brick doesn't make it obvious where errors are coming from
When brick fails due to errors in Lua, it outputs errors like this:

    ERR execution fail error="[string \"0cca71fcf76698693693c6edc703ea42c82494bfafa1b6a14100aabaa7a1afb08f\"]:27: expected string but got nil" cmd=call module=brick

This looks like a bug; presumably the `[string \"...\"]` bit is supposed to be a line number and 27 is the column number?  This makes it a lot harder to debug problems in Lua.

kroggen:
Fixes #113

### DIFF
--- a/contract/vm.go
+++ b/contract/vm.go
@@ -1340,7 +1340,7 @@ func vmAutoload(L *LState, funcName string) bool {
 }
 
 func (ce *executor) vmLoadCode(id []byte) {
-	address := C.CString(types.EncodeAddress(id))
+	address := C.CString("@" + types.EncodeAddress(id))
 	defer C.free(unsafe.Pointer(address))
 	if cErrMsg := C.vm_loadbuff(
 		ce.L,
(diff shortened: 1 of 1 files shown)


kroggen:
This changes the error messages in Lua from this:

```
"[string \"5a54b761f0908e98698db696a8623c910875a54b761f0908e98698db696a8623c9\"]:2: attempt to concatenate local 'name' (a nil value)"
```

To this:

```
"Amhfxh8VUPEpUGHuxyQCLn7shbyJeiQ2jujwV3i21g6b6JhDSoxQ:2: attempt to concatenate local 'name' (a nil value)"
```

kroggen:
To handle hardfork, the code could be something like this:

```go
func (ce *executor) vmLoadCode(id []byte) {
  if HardforkConfig.IsVXFork(ce.ctx.blockInfo.No) {
    chunkId := C.CString("@" + types.EncodeAddress(id))
  } else {
    chunkId := C.CString(hex.EncodeToString(id))
  }
```

PR #169, opened 2022-05-25
Fix internal error on testmode

### LINKED ISSUE #168: Internal error on failed transaction on testmode
This is a bug that only happens when the `testmode` is enabled, and only on the first transaction of an account and only if it fails. On this case the engine generates an internal error when trying to reset the account, with error:

```
Internal error: fee is greater than balance
```

This occurs because when an account is created in `testmode` it is filled with an amount of tokens, but only on the new state, not the old state, as can be seen [here](https://github.com/aergoio/aergo/blob/f0f524fdc884208b0230015a4aa8377df4c50ff4/state/statedb.go#L273).

Then when the `resetAccount()` is called on the failed txn, it calls Reset on the account, that in turn [replace the new state by the old state](https://github.com/aergoio/aergo/blob/f0f524fdc884208b0230015a4aa8377df4c50ff4/state/statedb.go#L240), but the old state has no funds, leading to the error later.

kroggen:
Fixes #168

### DIFF
--- a/state/statedb.go
+++ b/state/statedb.go
@@ -270,7 +270,7 @@ func (states *StateDB) GetAccountStateV(id []byte) (*V, error) {
 				sdb:    states,
 				id:     id,
 				aid:    aid,
-				oldV:   &types.State{},
+				oldV:   &types.State{Balance: amount.Bytes()},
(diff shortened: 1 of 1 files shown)


hayarobi (approved):
lgtm

PR #173, opened 2022-06-16
brick: multi-line commands on batch mode

kroggen:
These changes add support for multi-line commands on batch files

The commands can be break into many lines when inside back-sticks (`)

Example:

```
call ac1 0 contract set_values `[
 123,
 ["first","second","last"],
 ["and another one", true]
]` `"TEST"`
```


hayarobi (approved):
lgtm

PR #174, opened 2022-06-16
brick: check result on call command

### LINKED ISSUE #104: brick doesn't allow checking return value of contract calls
In brick, [`query`](https://github.com/aergoio/aergo/tree/master/cmd/brick#user-content-query) allows an optional `expected_query_result` parameter for checking the return value from the query.  However there is no equivalent parameter for checking the return value from a [`call`](https://github.com/aergoio/aergo/tree/master/cmd/brick#user-content-call).  This would be a useful feature.

kroggen:
Fixes #104


hayarobi:
I reviewed the code and willing to approve and merge this, but the conflict occurred while rebasing branch. @kroggen could you resolve this conflict?

kroggen:
The conflict was solved

PR #177, opened 2022-06-24
Amounts in decimal format

kroggen:
Currently some functions on smart contracts accept amounts with unit, like:

```
1 aergo
15 aergo
200 aergo
```

But when we specify a decimal amount like "1.5 aergo", they are not recognized.

**This pull request adds support for amounts in decimal format to the Aergo engine**

So they could also support this format:

```
0.021 aergo
0.5 aergo
13.9 aergo
```

The amounts must contain the "aergo" unit. Decimal format is not valid for gaer and aer

The above change applies to these functions inside smart contracts:
```
contract.send()
contract.call.value()
contract.deploy()
```
As well as these 3 commands in Composable Transactions:
```
call-send
pcall-send
send
```
And also to these commands on brick:
```
inject
getstate
send
call
deploy
```


hayarobi (approved):
lgtm

Issue #149, opened 2022-01-31
Proposal: Composable Transactions


kroggen:
## COMPOSABLE TRANSACTIONS

It is a method to call many contracts in a single transaction.

All the process is **atomic**. If some of the calls fail, all the transaction is rolled back.

It is used in cases where the user desires to process many things at the same time atomically.

## Examples

**Example 1:** Purchase airplane tickets and hotel reservation from different contracts, only if both succeed. If some of the purchases fail, the other is reverted (the user does not want one of them if the other is not acquired).

**Example 2:** Swap token and buy something with it. If the buy fails, the user does not want that token.

**Example 3:** Add liquidity to a pool (2 tokens) in a single transaction

## Implementation

It may be possible to implement it with backwards compatibility with existing transactions, by just adding a new transaction type:

```
type: COMPOSABLE_CALL
recipient: 0x00
```

And then the call information is stored in the `payload` in JSON format.

It would be a list of calls processed in order

```
[
  <call 1>
  <call 2>
  <call 3>
]
```

Each call containing the contract address, function to call, and arguments:

```
[
  [contract, function, arg1, arg2, arg3...],
  [contract, function, arg1, arg2, arg3...],
  [contract, function, arg1, arg2, arg3...]
]
```

### Example uses:

Buy 2 products or services on a single transaction:

```
[
  ["<token 1>","transfer","<amount 1>","<shop 1>","buy","product 1"],
  ["<token 2>","transfer","<amount 2>","<shop 2>","buy","product 2"],
]
```

Acquire token2 via swap and buy a product/service with it:

```
[
  ["<token 1>","transfer","<amount 1>","<swap pair>","swap",{"exact_output":"<amount 2>"}],
  ["<token 2>","transfer","<amount 2>","<shop>","buy","product"],
]
```

Add liquidity to a swap pair:

```
[
  ["<token 1>","transfer","<amount 1>","<swap_pair>","store_token"],
  ["<token 2>","transfer","<amount 2>","<swap_pair>","add_liquidity"],
]
```

## Native Aergo Tokens

The native aergo tokens can be used only on the first called function (it must be `payable` if the aergo amount is > 0). For the other functions we can use WAERGO.

One method is to "wrap" the aergo tokens on the first call and then use the WAERGO on the other calls:

```
type: COMPOSABLE_CALL
recipient: 0x00
amount: 1500000000000000000
payload: [
  ["<waergo>","wrap"],
[...]

PR #182, opened 2022-08-16
Composable Transactions

kroggen:
This implements the feature of [Composable Transactions](https://github.com/aergoio/aergo/wiki/Composable-Transactions)


gokch:
It seems able to make 'multicall contract', which call multiple contract in contract. Is there a difference between multiple contracts in a transaction and multiple contracts in a contract?

gokch:
composable transaction can execute multiple signatures atomically, but it also can by multisig in one transaction.

kroggen:
> It seems able to make 'multicall contract', which call multiple contract in contract. Is there a difference between multiple contracts in a transaction and multiple contracts in a contract?

I honestly cannot understand what you wrote

> and also, i suspect there isn`t rollback state appropriately when composable transaction failed.

I will add failure cases to the unit tests

> composable transaction can execute multiple signatures atomically, but it also can by multisig in one transaction.

what do you mean by "signatures"?

what is "multisig in one transaction"?

what do you want to say with the above sentence?

kroggen:
For reference:

https://github.com/aergoio/aergo/wiki/Composable-Transactions

gokch:
> I honestly cannot understand what you wrote
I want to ask 'why composable transaction need?', if it can be replaced with contract call.

> and also, i suspect there isn`t rollback state appropriately when composable transaction failed.
is it tested with governance like votingPowerRank?

hayarobi:
It seems like that there added another smart contract language, since composible transaction has its own commands and syntax. 
The functions listed in [the wiki](https://github.com/aergoio/aergo/wiki/Composable-Transactions) can be coded in lua contract. I think that making a lua contract that calls multiple contracts sequencially, because contract developers need not to learn new syntaxes and commands. One drawback of this approch is that lua contract must be deployed before calling, and this seems critical. 
How about add feature of ONE SHOT contract, which does not store deployed contract in statedb but just instantly run code single time. This works similar with composible transaction.

kroggen:
It appears that you both have not read the description on the Wiki

You can also check #149 

There is no deployment of new contract. This adds the feature to process many calls in a single transaction.

This feature is meant to be used by EOA (simple accounts) not by contracts

There is a new tx type `MULTICALL` for this purpose

The JSON script is added in the payload of a transaction, then it is processed by the Lua code in `vm_multicall.go`

Governance is not supported, so there is no need to check voting power

I am perceiving now that integration tests are lacking (they did not exist when this feature was created). I will add some

hayarobi:
I suggested a new feature like yours, after reading and inspired by that wiki .
The key feature is executing multiple contract call in a single transaction, atomically. And, some minor features. 
If the Argo node adds feature to immediately execute the Lua code inside the transaction payload (, but not deploying it), it can cover virtually all of the functions of composable transaction. 

Let say this types of transaction as `INSTANTCALL`, the payload of transaction contains aergo lua code. If aergo node receives this transaction, the node executes the `default` function inside that payload only once, with slightly different behavior from `CALL` transaction. The node does not deploy the code to state db nor store state values or global values. I said this as ONE SHOT. 

Your composable transaction looks quite simple and intuitive if it has a sequence of contract calls only, but the complexity looks significantly increased if it added result variables, scripts or some commands. I felt that it is another programming language.

kroggen:
There are problems when allowing smart contract language on transactions:

1. It is harder to understand when reviewing the transaction on the wallet
2. There should be limits on what is allowed on EOA (not allowed examples: governance, events...)

The script logic limits what can be done, and makes it somewhat understandable.
If the logic shown on the wallet is not clear, just do not sign the transaction.

The Aergo's Ledger App already has support for MultiCall, as can be seen [here](https://github.com/aergoio/ledger-app-aergo/blob/master/src/selection.h#L80) and [here](https://github.com/aergoio/ledger-app-aergo/blob/master/unit-tests/test_tx_display.c#L1765)

It is not the simplest way, but it gives a lot of power to the dApps

A simpler method is being created for simpler cases, but it is more limited than this

hayarobi:
There is a trade-off between verbosity and readability. The composable tx lacks of comment, argument type checking or etc, and provided limited functionality to manipulate variables and flow controls. If #266 is approved and put plain lua code for `INSTANTCALL`, both node and human will be able to review that transaction easier.

It seems possible to use a specific system function set for INSTANTCALL by applying the method of using system_lib_v* variables for the V4 hard fork.

I found a feature on Aptos chain, called [Move Script](https://aptos.dev/move/move-on-aptos/scripts/) , which has similar with my suggestion. For reference, smart contract code deployed intot the chain is called Module.

kroggen:
If you want to create an alternative, create an issue or open an item in discussions. Then put all the details there of how it should work, including how a normal user would sign such transaction on a hardware wallet, and examples of code for some cases

kroggen:
Also note that this feature does not aim to be a fully functional scripting language.

It started very basic, with just multiple contract calls and assertion, and then it was enhanced with some basic flow control.

It can also be released with just the basic commands.

Argument type checking is done by existing Lua functions. If an argument has unexpected type, it will fail and the whole transaction will be reverted.

shepelt:
I think it would make sense to enable composable transaction as a feature that can be enabled and deploy on alphanet to perform more extensive test on a live environment (same as SQL feature.)
That way, it can be used for private/enterprise environment as well.

kroggen:
That makes sense, it is a good option

kroggen:
[Text-Based Transactions](https://github.com/aergoio/aergo/issues/252) are an alternative but they do not cover all the use cases possible with Composable Transactions

Text-Based Transactions are better for:

* Token transfers (single or multiple, fungible and non-fungible)
* Token swaps (fungible and non-fungible)

And these are one of the most common use cases on a blockchain.

Composable Transactions cover the remaining cases, as it is more generic. They can be used with any kind of contract, not only those that deal with tokens.

Both above allow review of the commands on the hardware wallet, so the user can check before signing the transaction.

Making a token swap using Composable Transactions is indeed very verbose. Text-Based Transactions come to fix that.

hayarobi (approved):
lgtm

PR #193, opened 2023-05-03
add system.toAddress() and system.toPubkey()

### LINKED ISSUE #142: Smart contracts: convert between address and public key
Suggestion for smart contracts: a function (or 2) to convert between an aergo address and the associated public key

```
address <-> pubkey
```

This is useful when we have content that was signed outside of Aergo (other networks or apps) and so we have only the public key, but we need the corresponding aergo address to verify the signature

kroggen:
Closes #142 

`system.toPubkey(address)`: convert an address to a public key
`system.toAddress(pubkey)`: convert a public key to an address


kroggen:
TODO:

- [ ] add tests

kroggen:
It is giving this error message when passing a `nil`:

```
bad argument #1 to '?' (string expected, got nil)
```

Instead of `?` it should display the function name...

hayarobi (approved):
lgtm

PR #197, opened 2023-05-08
aergocli: deploy contract in hex format

kroggen:
Allow `aergocli` to accept contract code encoded in hex format.

This is used to deploy a copy of an existing contract to another chain, or even the same. Also used for test purposes, when deployed locally.

The data in hex format is copied from aergoscan view of the transaction that deployed the contract.

With this change, the payload given to `aergocli` on `contract deploy` can be either in hex format or the format generated by `aergoluac --payload`

### DIFF
--- a/cmd/aergocli/cmd/contract.go
+++ b/cmd/aergocli/cmd/contract.go
@@ -130,6 +130,11 @@ func init() {
 }
 
 func isHexString(s string) bool {
+  // check is the input has even number of characters
+  if len(s)%2 != 0 {
+    return false
+  }
+  // check if the input contains only hex characters
 	for _, c := range s {
 		if !((c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F')) {
 			return false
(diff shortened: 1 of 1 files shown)


cmd/aergocli/cmd/contract.go
```diff
@@ -128,6 +129,15 @@ func init() {
 	rootCmd.AddCommand(contractCmd)
 }
 
+func isHexString(s string) bool {
+	for _, c := range s {
```
hayarobi, line 133: It can be more safe by checking the length of string is even

kroggen: But the output of `aergoluac` can also be even...

kroggen: Oh, do you mean only inside the `isHexString()` function? Makes sense

kroggen:
Should the base be changed to `develop`?

hayarobi:
I think this PR should be merged to develop for now.
Anyway, it looks safe to merge develop into release since the develop doesn't have any changes that affecf behaviors of aergosvr binary, so I want to merge develop and release/2.4.4 if you agree.

kroggen:
OK, I changed the base to `develop`. You can make a new review or merge this.

It triggered new test.

I suggest to merge the other PR I sent too (only tests)

Then later you can merge develop into release/2.4.4

hayarobi (approved):
lgtm

PR #199, opened 2023-05-08
revert changes on failed pcall()

kroggen:
This PR changes/updates the behavior of both `pcall` or `xpcall` functions

When the called function fails, it will:

* Discard events emitted by the called functions and contracts
* Rollback the contract state (state variables, nonce, balance, sql db)


kroggen:
TODO: check `xpcall()`

hayarobi (approved):
lgtm

PR #203, opened 2023-05-10
[CI/CD] create build and test action

kroggen:


kroggen:
Changes to the `master` and `develop` branches must be done via pull requests, and the above action will run and must pass before any pull request is merged

kroggen:
There is no need to merge the `develop` into `release/2.4.4` after this is merged

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,51 @@
...
+      - reopened
+      - synchronize
+  push:
+    branches:
+      - master
+      - develop
+
+jobs:
+  build_and_test:
+    runs-on: ubuntu-latest
```
hayarobi, line 20: Is this means that test will run on different ubuntu version depending on the testing time?

kroggen: It will run on the last version of ubuntu available for actions

kroggen: One test run passed with the current config:

https://github.com/aergoio/aergo/actions/runs/4931682161/jobs/8813966127

kroggen:
It uses `go test -v ./...` instead of `make check` so we can see which specific test failed

PR #205, opened 2023-05-10
Topic/build libgmp

hayarobi:
This patch is intended to make building projects and working with git easier.
It removes source directory of libgmp-6.1.2 but adds tarball archive of libgmp which will be extracted in build time.

Removing libgmp source files make commands like `git ls-files` shows more meaningful output. It also makes it easy to change the version of libgmp, for future use. 
Also, I made it possible to build by selecting the libgmp-6.2.1 version. With this change, we improved the convenience of development by providing an option to compile the Aergo server on Apple Silicon Macs. (However, since Aergo's lua vm is architecture-dependent, it will be impossible to synchronize with Mainnet and Testnet nodes in Apple Silicon.)


hayarobi:
To build on Apple Silicon, you need to set environment variable `GMP_VERSION` to `6.2.1`.

kroggen:
Did you tested this branch on a MacBook M1/M2? Is it passing the tests there?

kroggen:
> To build on Apple Silicon, you need to set environment variable `GMP_VERSION` to `6.2.1`.

Maybe this can be done in the Makefile, detecting the OS and setting the environment variable

kroggen:
This is not tested

I suspect it should be put at the beginning of the Makefile, before calling cmake

```makefile
UNAME_S := $(shell uname -s)
ifeq ($(UNAME_S),Darwin)
  # It is running on MacOS
  export GMP_VERSION=6.2.1
endif
```

hayarobi:
I build on apple silicon mac like below;

`export GMP_VERSION=6.2.1 && make`

libtool/.gitignore
```diff
@@ -2,3 +2,7 @@ bin/
 include/
 lib/
 share/
+src/gmp-*
+/CMakeFiles
+/Makefile
+/cmake_install.cmake
```
kroggen, line 8: Is it correct to add a slash `/` at the beginning? Will it affect the Makefile at the root?

hayarobi: The leading slash means that this only apply to current directoy. `Makefile` in project root was already in repository and I want to ignore these files just in libtool directory.

kroggen:
The `make clean` command is failing:

```
~/aergo2# make clean
-- Using libgmp version to 6.1.2
-- Configuring done
-- Generating done
-- Build files have been written to: /root/aergo2/build
make[5]: *** No rule to make target 'clean'.  Stop.
make[4]: *** [libtool/CMakeFiles/libgmp-clean.dir/build.make:70: libtool/CMakeFiles/libgmp-clean] Error 2
make[3]: *** [CMakeFiles/Makefile2:646: libtool/CMakeFiles/libgmp-clean.dir/all] Error 2
make[2]: *** [CMakeFiles/Makefile2:389: CMakeFiles/distclean.dir/rule] Error 2
make[1]: *** [Makefile:254: distclean] Error 2
make: *** [Makefile:51: clean] Error 2
```

kroggen:
Also please rename this:

> Using libgmp version to 6.1.2

To this:  (remove the "to ")

> Using libgmp version 6.1.2

libtool/.gitignore
kroggen: Every place that I search says that a leading slash means the root folder, so it is better to remove the `/` from the beginning

kroggen: OK, I was reading wrong docs

hayarobi:
> The `make clean` command is failing:

I know, but I'm still new to cmake and haven't found a proper workaround. Anyway, `make clean` doesn't work properly too often, even on existing branches, so I figured it didn't make much difference and not so important now. I'll try to find a way to fix it next time.

kroggen:
OK, I agree.

Overall the build and test is working. I also tested on Alpine and it is OK.

This can be merged if yo wish

kroggen (approved):

PR #206, opened 2023-05-11
Topic/revert bignum func

hayarobi:
It's PR just for running github action.
I don't intend to revert these commit right now.


kroggen (approved):
As the tests passed, we can remove these changes as they are not so important

PR #207, opened 2023-05-18
Add github actions for unit tests

hayarobi:
- Any push of any branches will trigger short unit tests.
- User can trigger normal unit tests on demand.
- Also some long running unit tests will be skipped in short unit tests

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -1,9 +1,7 @@
 # This workflow will build a golang project
 # For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-go
 
-name: Integration Test
-
-run-name: Integration tests and unit tests of ${{ github.ref_type }} "${{ github.ref_name }}"
+name: Complete Test
 
 on:
   workflow_dispatch:
--- a/.github/workflows/default_unittest.yml
+++ b/.github/workflows/default_unittest.yml
@@ -3,8 +3,6 @@
 
 name:  Unit Tests
 
-run-name: Unit tests of ${{ github.ref_type }} "${{ github.ref_name }}" by @${{ github.actor }}
-
 on:
   workflow_dispatch:
 
--- a/.github/workflows/default_unittest.yml
+++ b/.github/workflows/default_unittest.yml
@@ -1,27 +1,15 @@
 # This workflow will build a golang project
 # For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-go
 
-name: Integration Test
+name:  Unit Tests
 
-run-name: Integration tests and unit tests of ${{ github.ref_type }} "${{ github.ref_name }}"
+run-name: Unit tests of ${{ github.ref_type }} "${{ github.ref_name }}" by @${{ github.actor }}
 
 on:
   workflow_dispatch:
-  pull_request:
-    types:
-      - opened
-      - reopened
-      - synchronize
-  pull_request_target:
-    types:
-      - edited
-  push:
-    branches:
-      - master
-      - develop
 
 jobs:
-  build_and_test:
+  build_and_unittest:
     runs-on: ubuntu-latest
 
     steps:
@@ -44,13 +32,4 @@ jobs:
       run: make
 
     - name: Unit Tests
-      run: go test -v ./...
-
-    - name: Integration Tests
-      run: |
-        if [ -d "tests" ]; then
-          cd tests
-          ./run_tests.sh
-        else
-          echo "The 'tests' folder does not exist."
-        fi
+      run: go test -timeout 999s -v ./...
--- a/.github/workflows/short_unittest.yml
+++ b/.github/workflows/short_unittest.yml
@@ -1,27 +1,18 @@
 # This workflow will build a golang project
 # For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-go
 
-name: Integration Test
+name:  Short Unit Tests
 
-run-name: Integration tests and unit tests of ${{ github.ref_type }} "${{ github.ref_name }}"
+run-name: Short unit tests of ${{ github.ref_type }} "${{ github.ref_name }}" by @${{ github.actor }}
 
 on:
-  workflow_dispatch:
-  pull_request:
-    types:
-      - opened
-      - reopened
-      - synchronize
-  pull_request_target:
-    types:
-      - edited
   push:
-    branches:
+    branches-ignore: # master and develop run full test by other workflow
       - master
       - develop
 
 jobs:
-  build_and_test:
+  build_and_unittest:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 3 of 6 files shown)


kroggen:
When this PR was created, it run 2 different checks. This is not needed. It should run only the main one, made for PRs

.github/workflows/default_unittest.yml
```diff
@@ -0,0 +1,35 @@
+# This workflow will build a golang project
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-go
+
+name:  Unit Tests
+
+run-name: Unit tests of ${{ github.ref_type }} "${{ github.ref_name }}" by @${{ github.actor }}
```
kroggen, line 6: This information is redundant. The list of actions already has information about user and branch

Check [here](https://github.com/aergoio/aergo/actions)

.github/workflows/short_unittest.yml
```diff
@@ -0,0 +1,38 @@
+# This workflow will build a golang project
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-go
+
+name:  Short Unit Tests
+
+run-name: Short unit tests of ${{ github.ref_type }} "${{ github.ref_name }}" by @${{ github.actor }}
```
kroggen, line 6: same

.github/workflows/short_unittest.yml
```diff
@@ -0,0 +1,38 @@
...
+
+on:
+  push:
+    branches-ignore: # master and develop run full test by other workflow
+      - master
+      - develop
```
kroggen, line 12: something must be added to prevent the run when a new pull request is created

.github/workflows/short_unittest.yml
```diff
@@ -0,0 +1,38 @@
...
+        go-version: 1.19
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get install -y jq m4 make cmake
+
+    - name: Build
+      run: make
+
+    - name: Unit Tests short
```
kroggen, line 37: Unit Tests (short)
or
Short Unit Tests

syncer/syncerservice_test.go
```diff
@@ -10,7 +11,14 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestMain(m *testing.M) {
+	os.Exit(m.Run())
+}
```
kroggen, line 16: What does this code do?

kroggen (changes requested):

syncer/syncerservice_test.go
hayarobi: It's a sign of incomplete cleanup. I initially tried to change the log level because the test was taking a lot of meaningless logs, but I gave up because it couldn't be done easily. I reverted that code and missed it. 

I'll fix those you said next week.

.github/workflows/build_and_test.yml
```diff
@@ -1,7 +1,9 @@
 # This workflow will build a golang project
 # For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-go
 
-name: Build and Test
+name: Integration Test
+
+run-name: Integration tests and unit tests of ${{ github.ref_type }} "${{ github.ref_name }}"
```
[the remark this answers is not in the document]
hayarobi: I've struggled with naming these two unit test workflows. 
Do you have any suggestions of good names that would easily distinguish between the following three?
1. short unit tests 
2. all unit tests
3. all unit tests and integration tests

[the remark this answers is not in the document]
kroggen: I mean the `run-name` can be removed

The `name` can be kept "Build and Test" or "Complete Tests"

"Integration Tests" is when we use `aergosvr` and `aergocli` together to check how they work integrated

We can also have integration tests with herajs, herapy, heraj, libaergo... together with `aergosvr` in test mode

And then when we have many `aergosvr` running on separate nodes, and test the deploy with the IDE and calls with wallets, then it is "Functional Test"

.github/workflows/default_unittest.yml
```diff
@@ -0,0 +1,33 @@
+# This workflow will build a golang project
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-go
+
+name:  Unit Tests
+
+on:
+  workflow_dispatch:
```
kroggen, line 7: is this file really needed? the complete test also can be triggered manually

.github/workflows/short_unittest.yml
```diff
@@ -0,0 +1,39 @@
...
+on:
+  push:
+    branches-ignore: # master and develop run full test by other workflow
+      - master
+      - develop
+
+jobs:
+  build_and_unittest:
+    # this prevents the action to run on pull requests
+    if: github.event_name == 'push' && github.ref != 'refs/pull/*/merge'
```
kroggen, line 15: This is not working

kroggen: This is an option: https://github.com/marketplace/actions/skip-duplicate-actions

kroggen:
Another option to avoid the duplicate run of actions is to unify the 2 actions into a single one.

First by removing the filter of branches on push on `build_and_test.yaml`

Then run the steps conditionally:

```yaml
    - name: Complete Unit Test
      if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/develop'
      run: ...

    - name: Short Unit Test
      if: github.ref != 'refs/heads/master' && github.ref != 'refs/heads/develop'
      run: ...
```

But I suspect that it will still run the action again when the PR is merged...

If this is the case, then [this](https://github.com/marketplace/actions/skip-duplicate-actions) would be better

.github/workflows/default_unittest.yml
hayarobi: I thought it depends on the size of integration tests. If running scripts `tests/run_tests.sh` takes much longer than whole unit tests. But after hearing what you said, I think it's better run it directly on the local machine when I want to test.

.github/workflows/short_unittest.yml
kroggen: The reason why the 2nd action did not run (was skipped) is because it had run before, not because the first action has run.

We can see this [here](https://github.com/aergoio/aergo/actions/runs/5077771284/jobs/9121449880?pr=207) where it states:

```
Skip execution because the exact same files have been successfully checked in run https://github.com/aergoio/aergo/actions/runs/5077707153
```

But that run is from the same action

kroggen: The other option bellow (joining them into a single job) may be better, and simpler

kroggen: I will try to implement and check the result

kroggen:
I made a test on another fork and it did not work

You can see the PR [here](https://github.com/kroggen/aergo/pull/2) it run the short unit tests instead of the complete tests on the pull request

kroggen (approved):
OK, it can be merged as is

kroggen:
If you will no more use the branch, then it can be deleted

PR #208, opened 2023-05-18
Tune unit test in vm_test.go

hayarobi:
- skip tests in vm_test.go on non amd64 architecture since those tests are archtecture dependent.

### DIFF
--- a/contract/vm_test.go
+++ b/contract/vm_test.go
@@ -8,12 +8,21 @@ import (
 	"io/ioutil"
 	"math/big"
 	"os"
+	"runtime"
 	"strconv"
 	"strings"
 	"testing"
 
 	"github.com/aergoio/aergo/types"
 )
+
+func TestMain(m *testing.M) {
+	if runtime.GOARCH != "amd64" {
+		fmt.Println("skip architecture dependent test")
+	} else {
+		os.Exit(m.Run())
+	}
+}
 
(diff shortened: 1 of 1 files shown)


contract/vm_test.go
```diff
@@ -8,13 +8,22 @@ import (
...
 	"strconv"
 	"strings"
 	"testing"
 
 	"github.com/aergoio/aergo/types"
 )
 
+func TestMain(m *testing.M) {
+	if runtime.GOARCH != "amd64" {
+		fmt.Println("skip architecture dependent test")
```
kroggen, line 21: It is just printing this, without command to skip
Is this working?

contract/vm_test.go
```diff
@@ -8,13 +8,22 @@ import (
...
+	} else {
+		os.Exit(m.Run())
```
kroggen, line 23: Is this not closing the test?

contract/vm_test.go
hayarobi: The code `m.Run()` do actual test, so no test will be run and I think it's actually same as skipping test. But it can show different message in `t.Skip()`. Is this can make a side effect?

kroggen: Oh, I understood. So there is no need to change it

kroggen (approved):

PR #210, opened 2023-05-31
publish coverage report

hayarobi:
This PR make two mutually exclusive `step`s; run unit tests as well as upload code coverage report to Code Climate in the case of master branch being pushed, or just run unit tests in other cases.
This structure can be somewhat confusing to users, but I failed to find a better way.


kroggen:
Did you test it on another fork to check if it is working?

hayarobi:
> Did you test it on another fork to check if it is working?

It will be better way than I did, I'll check more on the fork. I tested in using other branches on aergo and on other repository.

hayarobi:
Checked on another fork.

https://codeclimate.com/github/hayarobi/aergo

PR #214, opened 2023-06-13
brick: fix redefined flag

kroggen:
Fix error "redefined flag: v" on `brick`


kroggen:
@hayarobi 

I guess this started happening when the master and develop branches were merged

The `v` flag is already being used in the `log` module, so it gives the error "flag redefined: v" when we run `brick` 

I just renamed the flag to `b` on brick

Do you have any other idea?

hayarobi:
How about using capital `V`?

hayarobi (approved):
lgtm

kroggen:
Good idea! I changed it to `V`

As you can see, both the full and short workflows were triggered with this new commit. Only the full one should run. I made a fix to this in another PR

PR #215, opened 2023-06-14
fix github actions short test workflow

kroggen:
do not run short test on pull requests


kroggen:
This change fix the problem of Short Tests running on pull requests

The previous method was not working when a new commit/push is added to an existing pull request

hayarobi:
In my code, if the pushed commit had the same hash, the build and test will be skipped. Your change excludes short tests in case of pull request because it performs complete tests. These two `jobs` has somewhat different purpose. Is there any way to combine both?

kroggen:
For which cases the current method would be useful?

I suspect that if the commits are rebased or cherry-picked to another branch, the hash will be different, right?

hayarobi (approved):
lgtm

PR #218, opened 2023-06-16
Fix typos in chain and consusus packages

hayarobi:
Fixed a typo found while investigating the hardfork bug. I separated this typo-fix PRs from bug-fix PRs, which will be pushed later. it can make harder to tell what the logic change is if they're lumped together.


kroggen (approved):

hayarobi:
It looks like test vm_dummy was failed by timeout.

 FAIL	github.com/aergoio/aergo/contract/vm_dummy	999.015s

kroggen:
> It looks like test vm_dummy was failed by timeout.

It may be fixed by [this PR](https://github.com/aergoio/aergo/pull/220)

So both can be merged into develop. This here is already approved.

PR #219, opened 2023-06-16
add comments to code

kroggen:
And also rename some functions and variables to more clear names

### DIFF
--- a/contract/contract.go
+++ b/contract/contract.go
@@ -31,7 +31,7 @@ type preloadReply struct {
 }
 
 type preloader struct {
-	requestedTx *types.Tx
+	requestedTx *types.Tx  // the next preload tx to be executed
 	replyCh     chan *preloadReply
 }
 
@@ -60,6 +60,7 @@ func init() {
 	go preloadWorker()
 }
 
+// mark the next preload tx to be executed
 func SetPreloadTx(tx *types.Tx, service int) {
 	preloaders[service].requestedTx = tx
 }
--- a/contract/contract.go
+++ b/contract/contract.go
@@ -109,6 +109,7 @@ func Execute(
 	var gasLimit uint64
 	if useGas(bi.ForkVersion) {
 		if isFeeDelegation {
+			// check if the contract has enough balance for fee
 			balance := new(big.Int).Sub(receiver.Balance(), usedFee)
 			gasLimit = fee.MaxGasLimit(balance, bs.GasPrice)
 			if gasLimit == 0 {
@@ -116,8 +117,10 @@ func Execute(
 				return
 			}
 		} else {
+			// read the gas limit from the tx
 			gasLimit = txBody.GetGasLimit()
 			if gasLimit == 0 {
+				// no gas limit specified, the limit is the sender's balance
 				balance := new(big.Int).Sub(sender.Balance(), usedFee)
 				gasLimit = fee.MaxGasLimit(balance, bs.GasPrice)
 				if gasLimit == 0 {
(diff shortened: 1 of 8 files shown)


contract/vm.go
```diff
@@ -825,57 +829,71 @@ func Call(
...
-			evs := ce.getEvents()
-			if evs != nil {
+			events := ce.getEvents()
+			if events != nil {
 				_, _ = ctx.traceFile.WriteString("[Event]\n")
-				for _, ev := range evs {
-					eb, _ := ev.MarshalJSON()
-					_, _ = ctx.traceFile.Write(eb)
+				for _, event := range events {
+					event_json, _ := event.MarshalJSON()
```
gokch, line 860: golang generally uses camelCase. ( some linter warn snake_case. ) I think jsonEvent is better (but event_json is no problem too)

contract/contract.go
```diff
@@ -14,28 +14,30 @@ import (
...
-	tx  *types.Tx
-	ex  *executor
-	err error
-}
+/* The preloadWorker optimizes the execution time by preloading the next transaction while executing the current transaction */
 
-type preLoadReq struct {
-	preLoadService int
+type preloadRequest struct {
+	preloadService int
```
gokch, line 20: it is better to change preloadServiceType. ( BlockFactory and ChainService )
it would good also that add specific type, not int.

```
type preloadServiceType int

const (
	BlockFactory preloadServiceType = iota
	ChainService
	MaxVmService
)
```

contract/vm_callback.go
```diff
@@ -1444,32 +1584,37 @@ func LuaGetDbSnapshot(service C.int) *C.char {
 	return C.CString(strconv.FormatUint(curContract.rp, 10))
 }
 
-func moveGas(L *LState, ctx *vmContext) {
+func setRemainingGas(L *LState, ctx *vmContext) {
```
gokch, line 1587: it is global function unlike getRemainingGas.
below would be better.

```
func (s *vmContext) setRemainingGas(L *LState) {
    ...
}

```

contract/vm.go
kroggen: OK

contract/contract.go
kroggen: Done

kroggen: I made a change on [this commit](https://github.com/aergoio/aergo/pull/219/commits/c0ff64b7210298590386490a704f8d10acd4b871) but it affects many places, included passing this variable to and from C code

Then I decided to let this PR be just about renaming and comments

You can check the errors here:

```
Error: contract/contract.go:194:4: cannot use preloadService (variable of type preloadServiceType) as type int in argument to NewVmContext
Error: contract/contract.go:330:4: cannot use request.preloadService (variable of type preloadServiceType) as type int in argument to NewVmContext
Error: contract/vm.go:147:19: cannot use ChainService (constant 1 of type preloadServiceType) as type int in assignment
Error: contract/vm.go:1161:12: cannot use ChainService + 1 (constant 2 of type preloadServiceType) as type int in assignment
Error: contract/vm_callback.go:1498:15: invalid operation: service < BlockFactory (mismatched types _Ctype_int and preloadServiceType)
Error: contract/vm_callback.go:1510:16: invalid operation: service != BlockFactory (mismatched types _Ctype_int and preloadServiceType)
```

kroggen:
I cancelled the short tests workflow because the full test was running. There is no reason to run both

hayarobi (approved):
OK

hayarobi (approved):

contract/contract.go
gokch: ok.

PR #223, opened 2023-06-23
fix hardfork version on child Lua state

kroggen:
Current code is using hardfork version 2 when a new contract is deployed by an existing contract, via `contract.deploy()`

This fix makes the new Lua state to have the same hardfork version as the parent Lua state


kroggen:
This change probably affects the gas computation on transactions that deploy a contract, like ARC1 and ARC2 factories, so a sync test is required

gokch (approved):
mainnet, testnet sync test is success for this change.

PR #225, opened 2023-06-26
add stdint in vm.h

gokch:
## What is changed?
add stdint.h in vm.h.
window machine cant build aergo because it use mingw compiler.
( mingw don`t know size of uint8_t or uint16_t without import stdint. )

## Is there risk of fork?
unlikely happens.

### DIFF
--- a/contract/vm.h
+++ b/contract/vm.h
@@ -5,6 +5,7 @@
 #include <lauxlib.h>
 #include <luajit.h>
 #include <stdbool.h>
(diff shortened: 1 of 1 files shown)


hayarobi (approved):
the stdint.h is included by transitive dependencies. The command `gcc -H contract/vm.c` will show it. 
I think that this change will not change binary on Linux or macOS.

kroggen (approved):

PR #228, opened 2023-07-03
reset txn id on vm_dummy

kroggen:


kroggen:
The short test was skipped because the full test would be run

gokch (approved):

PR #233, opened 2023-07-07
vm_dummy: rename variables

kroggen:
this is to apply some consistency with tx element names:

sender, recipient, amount, payload


kroggen:
I skipped one test that was triggered due to editing PR description

PR #234, opened 2023-07-10
update luajit

kroggen:
The main purpose is to remove warnings from LuaJIT build, and have a cleaner build of aergo


kroggen:
If #237 is merged, then this one can be canceled

PR #236, opened 2023-07-18
replace panic to logger.Panic

gokch:
## What is changed?
print cause and explaination when an panic occurs.

## Is there risk of fork?
not at all. it is change only log message.


consensus/impl/raftv2/waldb.go
```diff
@@ -59,7 +59,8 @@ func (wal *WalDB) convertFromRaft(entries []raftpb.Entry) ([]*consensus.WalEntry
 		case raftpb.EntryConfChange:
 			return consensus.EntryConfChange
 		default:
-			panic("not support raftpb entrytype")
+			logger.Panic().Str("entry", types.RaftEntryToString(entry)).Msg("invalid entry type")
+			panic("invalid entry type")
```
kroggen, line 63: maybe there is an additional line here

kroggen (approved):

PR #237, opened 2023-07-18
remove mutex on Lua states

kroggen:
The mutex on Lua states appears to be unnecessary

The problem of difference in gas was fixed on hardfork 3 on #180 

A sync test will be run on available networks


gokch (approved):

gokch:
sync test finished about this pull request

hayarobi (approved):
lgtm

PR #238, opened 2023-08-02
Change package path to follow versionng policy of go module

hayarobi:
This PR is to fix the problem that Argo is only registered up to 1.11.1 in the go package repository.
[https://pkg.go.dev/github.com/aergoio/aergo](url)

The package naming convention of the sub was modified in the way suggested by the golang standard, and the package longitude of the import division stored in each go source file was modified to match the changed naming convention.
Refer to the page below for naming conventions and procedures for modification.
[https://go.dev/doc/modules/major-version](https://go.dev/doc/modules/major-version)


kroggen (approved):
OK. Just wait the sync test to finnish to merge

PR #241, opened 2023-08-11
txns of type NORMAL should not call contracts

kroggen:
These changes are intended to be implemented on hardfork V4:

* Transactions with type `NORMAL` should **not** call smart contracts
* Transactions with type `TRANSFER` can **only** interface with smart contracts on these conditions:
    * The `amount` must be greater than `0`
    * The payload must be empty

To transfer native Aergo tokens to a smart contract for a function that is not named "default" or that requires arguments, we must use a transaction with type `CALL`


kroggen:
Tests to be added:

### On hardfork_version >= 4

Should work:

* Transfer Aergo to contract using `TRANSFER` type

Should fail:

* Transfer Aergo to contract using `TRANSFER` type, contract without function "default"
* Transfer Aergo to contract using `TRANSFER` type, contract function "default" not payable
* Transfer Aergo to contract using `TRANSFER` type, payload not empty (hash)
* Transfer Aergo to contract using `TRANSFER` type, payload not empty (JSON for call)
* Transfer Aergo to contract using `TRANSFER` type, amount = `0`

Then repeat the above using `NORMAL` type

### On hardfork_version < 4

Should fail:

* Transfer Aergo to contract using `TRANSFER` type, contract without function "default"
* Transfer Aergo to contract using `TRANSFER` type, contract function "default" not payable
* Transfer Aergo to contract using `TRANSFER` type, payload not empty (hash)

The same for `NORMAL` type

Should work:

* Transfer Aergo to contract using `TRANSFER` type
* Transfer Aergo to contract using `TRANSFER` type, payload not empty (JSON for call)
* Transfer Aergo to contract using `TRANSFER` type, amount = `0`

The same for `NORMAL` type

hayarobi:
@ultrafellen Do you agree with this policy?

ultrafellen (commented):
I'm not sure about the code, but I agree with the policy.

hayarobi (approved):
lgtm

PR #255, opened 2023-09-23
add contract.gasLeft()

kroggen:
This PR adds the `contract.gasLeft()` function to retrieve the remaining gas in the contract execution

There is a problem though:

The `lua_pushinteger()` function accepts a value of type `lua_Integer`, that is signed int64, and the `lua_gasget()` function returns an unsigned int64

So if the value of gas is too big, it may appear for the Lua code as a negative value (to be tested)

### DIFF
--- a/contract/contract_module.c
+++ b/contract/contract_module.c
@@ -281,6 +281,12 @@ static int moduleBalance(lua_State *L) {
 	return 1;
 }
 
+static int moduleGasLeft(lua_State *L) {
+	// do not charge gas for this call for easier calculations
+	lua_pushinteger(L, lua_gasget(L));
+	return 1;
+}
+
 static int modulePcall(lua_State *L) {
 	int argc = lua_gettop(L) - 1;
 	int service = getLuaExecContext(L);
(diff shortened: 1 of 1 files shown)


kroggen:
One solution is to use `bignum` for `gasLeft()`

But then other gas related functions should use `bignum` as well, for consistency

kroggen:
When the `gaslimit` is not set on the transaction, or when it is set to unlimited, the `gasLeft` is set to UINT64_MAX

So the `gaslimit` may not be so useful to detect if the txn is about to run out of gas due to the txn creator's balance being low

It may be useful only to limit gas usage on a called contract. But the method used for that would make the transaction to fail (and the changes be reverted). It would be useful to have a way to set a limit and do not revert the txn, like a protected call that just returns `false` if it runs out of gas

Maybe this PR do not need to be applied on V4, and can be left to the next hardfork

kroggen:
Maybe this PR do not need to be applied on V4, and can be left to the next hardfork

hayarobi:
In the current version, gas usage is increased on garbage collection. This means that it is virtually impossible for the user to determine the actual remaining gas since it is executed after the contract function call is returned. We need to change the way gas is calculated first, for this function to be meaningful.

kroggen:
This is changed on V4, [this commit](https://github.com/aergoio/luajit/commit/58772eda8070dd0cba167c28624626df9b79eedf) disables gas consumption for garbage collection

Although not tested

kroggen:
But the "remaining gas" is computed based on the transaction "gas limit", and if it is not specified, it is set to UINT64_MAX, even if it is higher than what the account's balance can pay.

The reason may be because the account can receive tokens while the transaction is executed, on one of the called contracts, so the execution is not limited on the initial balance of the account

It turns out that the number is huge and makes no much sense when used on `contract.gasLeft()`

Maybe this can be implemented later when the gas limit per block is implemented

PR #259, opened 2023-09-25
apply system params on next block

kroggen:
If enough votes to change a system param value is reached, save the new value on a temporary place and only update the param value when the block is finished

In this way the new value will only be active for the next block

Also keep the values from params on memory instead of loading params from state on every block


kroggen:
It would be useful to have some tests for this

gokch:
I add mutex about systemParams because it occur panic when access map in same time.

gokch:
it need additional sync test before merge.

contract/system/param.go
```diff
@@ -86,51 +134,45 @@ func updateParam(s dataSetter, id string, value *big.Int) (error) {
...
+	if val := systemParams.getParam(proposalID); val != nil {
 		return val
 	}
-	return DefaultParams[proposalID]
+	return DefaultParams.getParam(proposalID)
 }
 
 // these 4 functions are reading the param value for the current block
-
 func GetStakingMinimum() *big.Int {
```
kroggen, line 182: This comment is for the next 4 function, so it should have an empty line

contract/system/param.go
```diff
@@ -148,7 +190,6 @@ func GetBpCount() int {
 }
 
 // these functions are reading the param value directly from the state
-
 func GetNamePriceFromState(scs *state.ContractState) *big.Int {
```
kroggen, line 199: Same

contract/system/param.go
gokch: Is these our official convention? all code can be changed to using these convension.

gokch: personally, i prefer these convension to division line
image

kroggen: loogs good

kslee8282:
Start sync test (testnet)
Oct 13 05:00:30.592 INF ../go/aergo/cmd/aergosvr/aergosvr.go:100 > AERGO SVR STARTED branch=topic/system-params module=asvr revision=a282031c

kslee8282:
Start sync test (mainnet)
Oct 13 05:05:40.440 INF ../go/aergo/cmd/aergosvr/aergosvr.go:100 > AERGO SVR STARTED branch=topic/system-params module=asvr revision=a282031c

kslee8282:
Sync completed

kroggen:
I am thinking in renaming these functions:

* `getNextParam` -> `getNextBlockParam`
* `setNextParam` -> `setNextBlockParam`
* `delNextParam` -> `delNextBlockParam`

It does not require a new sync test

@hayarobi @gokch  What do you think?

contract/system/param.go
```diff
@@ -26,7 +70,10 @@ const (
 )
 
 var (
-	systemParams parameters
+	systemParams *parameters = &parameters{
+		mutex:  sync.RWMutex{},
+		params: map[string]*big.Int{},
+	}
```
kroggen, line 76: I suspect that the above can be just `systemParams *parameters`
It is loaded on `loadParams()`

gokch:
> I am thinking in renaming these functions:
> 
> * `getNextParam` -> `getNextBlockParam`
> * `setNextParam` -> `setNextBlockParam`
> * `delNextParam` -> `delNextBlockParam`
> 
> It does not require a new sync test
> 
> @hayarobi @gokch What do you think?

i agree.

kroggen:
Can this be merged into develop?

gokch (approved):

PR #262, opened 2023-09-25
AMAIN-176 Fix block timeout check

hayarobi:
This patch fixes a bug that causes synchronization failure in situations when there are too many transactions.

There were two bugs, actually.

First, when creating a block, if the transaction execution timeout (1/4 of the block creation interval = 0.25 seconds) is reached, execution of the remaining transactions must be abandoned and a block must be created with only completed transactions, but the code does not work properly and the remaining transactions continue to be executed, so the block is created after a long time. I've changed the logic that checks timeout with one global channel variable to the logic that checks timeout by using `context.Context` created per a block.

Another problem was that when a timeout occurred during the execution of a transaction of multi-depth contract function - calling other contract inside function, the failed transaction should be abandoned and evicted, but added in the block due to a bug. The timeout is determined by real time and non-deterministic. That timeout transaction was executed as success on verification node, so that the block hash became different with block producer.

### DIFF
--- a/chain/chainhandle.go
+++ b/chain/chainhandle.go
@@ -26,12 +26,12 @@ import (
 
 var (
 	ErrorNoAncestor      = errors.New("not found ancestor")
-	ErrBlockOrphan       = errors.New("block is ohphan, so not connected in chain")
+	ErrBlockOrphan       = errors.New("block is orphan, so not connected in chain")
 	ErrBlockCachedErrLRU = errors.New("block is in errored blocks cache")
 	ErrStateNoMarker     = errors.New("statedb marker of block is not exists")
 
 	errBlockStale       = errors.New("produced block becomes stale")
-	errBlockInvalidFork = errors.New("invalid fork occured")
+	errBlockInvalidFork = errors.New("invalid fork occurred")
 	errBlockTimestamp   = errors.New("invalid timestamp")
 
 	InAddBlock      = make(chan struct{}, 1)
@@ -384,7 +384,7 @@ func (cp *chainProcessor) reorganize() error {
 	return nil
 }
 
-func (cs *ChainService) addBlockInternal(newBlock *types.Block, usedBstate *state.BlockState, peerID types.PeerID) (err error, cache bool) {
+func (cs *ChainService) addBlockInternal(newBlock *types.Block, usedBState *state.BlockState, peerID types.PeerID) (err error, cache bool) {
 	if !cs.VerifyTimestamp(newBlock) {
 		return &ErrBlock{
 			err: errBlockTimestamp,
@@ -405,12 +405,12 @@ func (cs *ChainService) addBlockInternal(newBlock *types.Block, usedBstate *stat
 	}
 
 	// The newly produced block becomes stale because the more block(s) are
-	// connected to the blockchain so that the best block is cha/nged. In this
+	// connected to the blockchain so that the best block is changed. In this
 	// case, newBlock is rejected because it is unlikely that newBlock belongs
-	// to the main branch. Warning: the condition 'usedBstate != nil' is used
+	// to the main branch. Warning: the condition 'usedBState != nil' is used
 	// to check whether newBlock is produced by the current node itself. Later,
 	// more explicit condition may be needed instead of this.
-	if usedBstate != nil && newBlock.PrevID() != bestBlock.ID() {
+	if usedBState != nil && newBlock.PrevID() != bestBlock.ID() {
 		return &ErrBlock{
 			err: errBlockStale,
 			block: &types.BlockInfo{
@@ -425,7 +425,7 @@ func (cs *ChainService) addBlockInternal(newBlock *types.Block, usedBstate *stat
 		if cs.IsForkEnable() {
 			return nil
 		}
-		if usedBstate != nil {
+		if usedBState != nil {
 			return nil
 		}
 
@@ -457,7 +457,7 @@ func (cs *ChainService) addBlockInternal(newBlock *types.Block, usedBstate *stat
 
 	// handle orphan
 	if cs.isOrphan(newBlock) {
-		if usedBstate != nil {
+		if usedBState != nil {
 			return fmt.Errorf("block received from BP can not be orphan"), false
 		}
 		err := cs.handleOrphan(newBlock, bestBlock, peerID)
@@ -476,7 +476,7 @@ func (cs *ChainService) addBlockInternal(newBlock *types.Block, usedBstate *stat
 		<-InAddBlock
 	}()
 
-	cp, err := newChainProcessor(newBlock, usedBstate, cs)
+	cp, err := newChainProcessor(newBlock, usedBState, cs)
 	if err != nil {
 		return err, true
 	}
@@ -496,7 +496,7 @@ func (cs *ChainService) addBlockInternal(newBlock *types.Block, usedBstate *stat
 	return nil, true
 }
 
-func (cs *ChainService) addBlock(newBlock *types.Block, usedBstate *state.BlockState, peerID types.PeerID) error {
+func (cs *ChainService) addBlock(newBlock *types.Block, usedBState *state.BlockState, peerID types.PeerID) error {
 	hashID := types.ToHashID(newBlock.BlockHash())
 
 	if cs.errBlocks.Contains(hashID) {
@@ -511,7 +511,7 @@ func (cs *ChainService) addBlock(newBlock *types.Block, usedBstate *state.BlockS
 	}
 
 	var needCache bool
-	err, needCache = cs.addBlockInternal(newBlock, usedBstate, peerID)
+	err, needCache = cs.addBlockInternal(newBlock, usedBState, peerID)
 	if err != nil {
 		if needCache {
 			evicted := cs.errBlocks.Add(hashID, newBlock)
@@ -562,7 +562,7 @@ type blockExecutor struct {
 	execTx           TxExecFn
 	txs              []*types.Tx
 	validatePost     ValidatePostFn
-	coinbaseAcccount []byte
+	coinbaseAccount  []byte
 	commitOnly       bool
 	verifyOnly       bool
 	validateSignWait ValidateSignWaitFn
@@ -577,9 +577,9 @@ func newBlockExecutor(cs *ChainService, bState *state.BlockState, block *types.B
 	commitOnly := false
 
 	// The DPoS block factory executes transactions during block generation. In
-	// such a case it send block with block state so that bState != nil. On the
+	// such a case it sends block with block state so that bState != nil. On the
 	// contrary, the block propagated from the network is not half-executed.
-	// Hence we need a new block state and tx executor (execTx).
+	// Hence, we need a new block state and tx executor (execTx).
 	if bState == nil {
 		if err := cs.validator.ValidateBlock(block); err != nil {
 			return nil, err
@@ -605,11 +605,11 @@ func newBlockExecutor(cs *ChainService, bState *state.BlockState, block *types.B
 	bState.Receipts().SetHardFork(cs.cfg.Hardfork, block.BlockNo())
 
 	return &blockExecutor{
-		BlockState:       bState,
-		sdb:              cs.sdb,
-		execTx:           exec,
-		txs:              block.GetBody().GetTxs(),
-		coinbaseAcccount: block.GetHeader().GetCoinbaseAccount(),
+		BlockState:      bState,
+		sdb:             cs.sdb,
+		execTx:          exec,
+		txs:             block.GetBody().GetTxs(),
+		coinbaseAccount: block.GetHeader().GetCoinbaseAccount(),
 		validatePost: func() error {
 			return cs.validator.ValidatePost(bState.GetRoot(), bState.Receipts(), block)
 		},
@@ -674,8 +674,8 @@ func (e *blockExecutor) execute() error {
 			}
 		}
 
-		//TODO check result of verifing txs
-		if err := SendBlockReward(e.BlockState, e.coinbaseAcccount); err != nil {
+		//TODO check result of verifying txs
+		if err := SendBlockReward(e.BlockState, e.coinbaseAccount); err != nil {
 			return err
 		}
 
--- a/chain/chainhandle.go
+++ b/chain/chainhandle.go
@@ -689,6 +689,7 @@ func (e *blockExecutor) execute() error {
 	}
 
 	if err := e.validatePost(); err != nil {
+		// TODO write verbose tx result if debug log is enabled
 		return err
 	}
 
--- a/chain/chainhandle.go
+++ b/chain/chainhandle.go
@@ -591,7 +591,8 @@ func newBlockExecutor(cs *ChainService, bState *state.BlockState, block *types.B
 			state.SetPrevBlockHash(block.GetHeader().GetPrevBlockHash()),
 		)
 		bi = types.NewBlockHeaderInfo(block)
-		exec = NewTxExecutor(nil, cs.ChainConsensus, cs.cdb, bi, contract.ChainService)
+		// FIXME currently the verify only function is allowed long execution time,
+		exec = NewTxExecutor(context.Background(), cs.ChainConsensus, cs.cdb, bi, contract.ChainService)
 
 		validateSignWait = func() error {
 			return cs.validator.WaitVerifyDone()
@@ -622,7 +623,7 @@ func newBlockExecutor(cs *ChainService, bState *state.BlockState, block *types.B
 }
 
 // NewTxExecutor returns a new TxExecFn.
-func NewTxExecutor(ctx context.Context, ccc consensus.ChainConsensusCluster, cdb contract.ChainAccessor, bi *types.BlockHeaderInfo, preloadService int) TxExecFn {
+func NewTxExecutor(execCtx context.Context, ccc consensus.ChainConsensusCluster, cdb contract.ChainAccessor, bi *types.BlockHeaderInfo, preloadService int) TxExecFn {
 	return func(bState *state.BlockState, tx types.Transaction) error {
 		if bState == nil {
 			logger.Error().Msg("bstate is nil in txExec")
@@ -634,7 +635,7 @@ func NewTxExecutor(ctx context.Context, ccc consensus.ChainConsensusCluster, cdb
 		}
 		blockSnap := bState.Snapshot()
 
-		err := executeTx(ctx, ccc, cdb, bState, tx, bi, preloadService)
+		err := executeTx(execCtx, ccc, cdb, bState, tx, bi, preloadService)
 		if err != nil {
 			logger.Error().Err(err).Str("hash", enc.ToString(tx.GetHash())).Msg("tx failed")
 			if err2 := bState.Rollback(blockSnap); err2 != nil {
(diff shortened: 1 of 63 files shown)


chain/chainhandle.go
```diff
@@ -621,10 +622,10 @@ func newBlockExecutor(cs *ChainService, bState *state.BlockState, block *types.B
 }
 
 // NewTxExecutor returns a new TxExecFn.
-func NewTxExecutor(ccc consensus.ChainConsensusCluster, cdb contract.ChainAccessor, bi *types.BlockHeaderInfo, preloadService int) TxExecFn {
+func NewTxExecutor(ctx context.Context, ccc consensus.ChainConsensusCluster, cdb contract.ChainAccessor, bi *types.BlockHeaderInfo, preloadService int) TxExecFn {
```
kroggen, line 625: `ctx` -> `execCtx`, to avoid confusion with `ctx` from `vmContext`

contract/vm.go
```diff
@@ -52,6 +53,8 @@ const (
 	maxCallDepth         = 64
 	checkFeeDelegationFn = "check_delegation"
 	constructor          = "constructor"
+
+	vmTimeoutErrMsg = "contract timeout of VmTimeoutError"
```
kroggen, line 57: "of" or "or"?

it can also be aligned

contract/vm.go
```diff
@@ -93,7 +99,7 @@ type vmContext struct {
 	node              string
 	confirmed         bool
 	isQuery           bool
-	nestedView        int32
+	nestedView        int32 // nestedView indicate which parent called called the contract in view (read-only mode)
```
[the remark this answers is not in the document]
kroggen: The comment can be:
```
// indicates which parent called the contract in view (read-only) mode
```

contract/vm.go
```diff
@@ -235,7 +242,7 @@ func (ctx *vmContext) IsGasSystem() bool {
 }
 
 // get the remaining gas from the given LState
-func (ctx *vmContext) getRemainingGas(L *LState) {
+func (ctx *vmContext) refreshRemainingGas(L *LState) {
```
kroggen, line 245: this id undoing my changes
there is a pair of functions `getRemainingGas` and `setRemainingGas` that were just put together on #261

contract/vm.go
```diff
@@ -600,27 +607,31 @@ func (ce *executor) call(instLimit C.int, target *LState) C.int {
 	}
 	ce.processArgs()
 	if ce.err != nil {
-		ctrLgr.Debug().Err(ce.err).Str("contract",
-			types.EncodeAddress(ce.ctx.curContract.contractId)).Msg("invalid argument")
+		ctrLgr.Debug().Err(ce.err).Stringer("contract",
+			types.LogAddr(ce.ctx.curContract.contractId)).Msg("invalid argument")
```
kroggen, line 611: I don't understand why this change
What is the difference int the log output?

contract/vm.go
```diff
@@ -600,27 +607,31 @@ func (ce *executor) call(instLimit C.int, target *LState) C.int {
...
+	vmLogger.Trace().Int64("execµs", vmExecTime).Stringer("contract", types.LogAddr(ce.ctx.curContract.contractId)).Msg("contract execute time in vm")
+	if cErrMsg != nil {
 		errMsg := C.GoString(cErrMsg)
 		if C.luaL_hassyserror(ce.L) != C.int(0) {
 			ce.err = newVmSystemError(errors.New(errMsg))
 		} else {
-			if C.luaL_hasuncatchablerror(ce.L) != C.int(0) &&
-				C.ERR_BF_TIMEOUT == errMsg {
+			isUncatchable := C.luaL_hasuncatchablerror(ce.L) != C.int(0)
+			if isUncatchable && (C.ERR_BF_TIMEOUT == errMsg || vmTimeoutErrMsg == errMsg) {
```
kroggen, line 626: variable before the constant, for consistency:
```go
errMsg == C.ERR_BF_TIMEOUT || errMsg == vmTimeoutErrMsg
```

contract/vm.go
```diff
@@ -600,27 +607,31 @@ func (ce *executor) call(instLimit C.int, target *LState) C.int {
...
+			types.LogAddr(ce.ctx.curContract.contractId)).Msg("invalid argument")
 		return 0
 	}
 	ce.setCountHook(instLimit)
 	nret := C.int(0)
-	if cErrMsg := C.vm_pcall(ce.L, ce.numArgs, &nret); cErrMsg != nil {
+	startTime := time.Now()
+	cErrMsg := C.vm_pcall(ce.L, ce.numArgs, &nret)
+	vmExecTime := time.Now().Sub(startTime).Microseconds()
+	vmLogger.Trace().Int64("execµs", vmExecTime).Stringer("contract", types.LogAddr(ce.ctx.curContract.contractId)).Msg("contract execute time in vm")
```
kroggen, line 619: is this time measurement above still needed?

contract/vm.go
```diff
@@ -783,8 +794,8 @@ func (ce *executor) close() {
 	}
 }
 
-func (ce *executor) getRemainingGas() {
-	ce.ctx.getRemainingGas(ce.L)
+func (ce *executor) refreshRemainingGas() {
+	ce.ctx.refreshRemainingGas(ce.L)
```
kroggen, line 798: You can rename the one from the executor, but keep the one of vmContext unchanged because it forms a pair of functions with similar functionality and name

contract/vm.go
```diff
@@ -839,8 +850,11 @@ func Call(
 	ce := newExecutor(contract, contractAddress, ctx, &ci, ctx.curContract.amount, false, false, contractState)
 	defer ce.close()
 
+	startTime := time.Now()
 	// execute the contract call
 	ce.call(callMaxInstLimit, nil)
+	vmExecTime := time.Now().Sub(startTime).Microseconds()
+	vmLogger.Trace().Int64("execµs", vmExecTime).Stringer("txHash", types.LogBase58(ce.ctx.txHash)).Msg("tx execute time in vm")
```
kroggen, line 857: is this still needed? can it be enabled only on debug mode? or maybe just remove it

contract/vm_callback.go
```diff
@@ -1492,6 +1500,10 @@ func luaCheckView(service C.int) C.int {
 	return C.int(ctx.nestedView)
 }
 
+// luaCheckTimeout checks two types of timeouts;
+// One is whether the block creation timeout is exceeded, and the other is whether
+// the transaction execution timeout is exceeded.
```
kroggen, line 1505: ; -> :
is -> was (2 times)

kroggen: Or like this:
```
// luaCheckTimeout checks two types of timeouts:
// 1. whether the block creation timeout was exceeded
// 2. whether the transaction execution timeout was exceeded
```

contract/vm_callback.go
```diff
@@ -1510,28 +1522,13 @@ func luaCheckTimeout(service C.int) C.int {
 	if service != BlockFactory {
 		return 0
 	}
-
+	ctx := contexts[service]
```
kroggen, line 1525: an empty line above would be useful

consensus/impl/dpos/blockfactory.go
```diff
@@ -32,13 +33,15 @@ const (
 )
 
 type txExec struct {
+	ctx    context.Context
 	execTx bc.TxExecFn
 }
 
-func newTxExec(cdb contract.ChainAccessor, bi *types.BlockHeaderInfo) chain.TxOp {
+func newTxExec(ctx context.Context, cdb contract.ChainAccessor, bi *types.BlockHeaderInfo) chain.TxOp {
```
kroggen, line 40: rename ctx to execCtx?

consensus/impl/dpos/blockfactory.go
```diff
@@ -163,7 +173,7 @@ func (bf *BlockFactory) controller() {
 				continue
 			}
 
-			notifyBpTimeout(bpi)
+			//notifyBpTimeout(bpi)
```
kroggen, line 176: this can be removed

contract/errors.go
```diff
@@ -53,7 +53,7 @@ func (e *VmSystemError) System() bool {
 type VmTimeoutError struct{}
 
 func (e *VmTimeoutError) Error() string {
-	return "contract timeout"
+	return "contract timeout of VmTimeoutError"
```
kroggen, line 56: "of" or "or"?

contract/vm.go
```diff
@@ -106,6 +112,7 @@ type vmContext struct {
 	traceFile         *os.File
 	gasLimit          uint64
 	remainedGas       uint64
+	txContext         context.Context
```
kroggen, line 115: txContext -> execCtx?

contract/vm.go
```diff
@@ -171,9 +178,7 @@ func getTraceFile(blkno uint64, tx []byte) *os.File {
 	return f
 }
 
-func NewVmContext(blockState *state.BlockState, cdb ChainAccessor, sender, reciever *state.V,
-	contractState *state.ContractState, senderID []byte, txHash []byte, bi *types.BlockHeaderInfo, node string, confirmed bool,
-	query bool, rp uint64, service int, amount *big.Int, gasLimit uint64, feeDelegation bool) *vmContext {
+func NewVmContext(txContext context.Context, blockState *state.BlockState, cdb ChainAccessor, sender, reciever *state.V, contractState *state.ContractState, senderID, txHash []byte, bi *types.BlockHeaderInfo, node string, confirmed, query bool, rp uint64, service int, amount *big.Int, gasLimit uint64, feeDelegation bool) *vmContext {
```
kroggen, line 181: txContext -> execCtx?

kroggen:
In the file there are many names for what appears to be the same thing:

```
txCtx
txContext
execCtx
...
```

They could all use the same name for consistency

kroggen:
I suspect that there is a problem:

On V4 the smart contracts will be compiled by the aergo nodes, and some big contracts can take ~500ms to compile or even more

Some of these contracts are deployed directly by a transaction and others are deployed by another contract

We need to handle these long executions in some way

consensus/impl/dpos/blockfactory.go
```diff
@@ -124,13 +141,14 @@ func (bf *BlockFactory) controller() {
 			return err
 		}
 
-		timeLeft := bpi.slot.RemainingTimeMS()
-		if timeLeft <= 0 {
-			return chain.ErrTimeout{Kind: "slot", Timeout: timeLeft}
+		timeLeftMSs := bpi.slot.GetBpTimeout()
+		if timeLeftMSs <= 0 {
```
kroggen, line 145: Why `MSs`? If acronym for milliseconds, then it can be just `timeLeftMS`

consensus/impl/dpos/slot/slot.go
```diff
@@ -136,7 +136,7 @@ func (s *Slot) NextBpIndex(bpCount uint16) int64 {
 
 // BpMaxTime returns the max time limit for block production in nsec.
 func BpMaxTime() time.Duration {
-	return time.Duration(bpMaxTimeLimitMs) * 1000
+	return time.Duration(bpMaxTimeLimitMs) * time.Millisecond
```
kroggen, line 139: Is this correct? The comment says "nsec" and the function uses `time.Millisecond`...

contract/vm_callback.go
kroggen: Or use "occurred" instead:
```
// luaCheckTimeout checks two types of timeouts:
// 1. whether the block creation timeout occurred
// 2. whether the transaction execution timeout occurred
```

kroggen:
Some notes about this PR:

* The branch name could be more descriptive instead of using "AMAIN..."
* The fixes of variable names and comments could have been done in a separate PR. It would be already merged at this time, and this PR would be smaller, easier to navigate and review

gokch:
> I suspect that there is a problem:
> 
> On V4 the smart contracts will be compiled by the aergo nodes, and some big contracts can take ~500ms to compile or even more
> 
> Some of these contracts are deployed directly by a transaction and others are deployed by another contract
> 
> We need to handle these long executions in some way

it seems not related this pull request. block is created in 1 second in dpos, so it cannot be resolved to just change timeout logic.

kroggen:
Only 250ms left for transaction execution, including contract compilation... that is not sufficient

Discussion here: #264

contract/vm.go
hayarobi: I was not sure how to describe this error effectively. I will change the message soon.

kroggen: if this error message can appear on receipts, then it should not contain "VmTimeoutError"
if used only for logs, it is OK

contract/vm.go
hayarobi: This is one of the log optimization. The former calls the EncodeAddress function regardless of the log level, but the latter does not call the encode function if the log level is lower than Debug. The outputs are same.

contract/vm.go
hayarobi: The previous name looks not close to common naming convention, so I was confused when I firstly saw that function. A function of name starting with `get..` is supposed to return a value and not to change internal field or state, at least in Java. Are there different conventions in golang?

kroggen: OK, I agree. The comment that I added may be sufficient to understand what it is doing

contract/vm.go
```diff
@@ -376,7 +383,7 @@ func newExecutor(
 	if ctx.IsGasSystem() {
 		ce.setGas()
 		defer func() {
-			ce.getRemainingGas()
+			ce.refreshRemainingGas()
```
kroggen, line 386: as the name of function is not clear, a comment is needed here

hayarobi:
> Some notes about this PR:
> 
>     * The branch name could be more descriptive instead of using "AMAIN..."
> 
>     * The fixes of variable names and comments could have been done in a separate PR. It would be already merged at this time, and this PR would be smaller, easier to navigate and review
> 
> 
> It is better when a PR has only code related to its proposal

the branch name is linked to issue tracker that my team uses.

kroggen:
You should use the public issue tracker of this repo and use clear names for branches

And also use the public discussion more instead of Slack

contract/vm_callback.go
kroggen: You did not add the notes above. Is the function not checking 2 timeouts anymore?

contract/errors.go
kroggen: for consistency, it can be  "contract timeout during vm execution"

consensus/impl/dpos/slot/slot.go
```diff
@@ -136,6 +136,7 @@ func (s *Slot) NextBpIndex(bpCount uint16) int64 {
 
 // BpMaxTime returns the max time limit for block production in nsec.
 func BpMaxTime() time.Duration {
+	// FIXME this function has a bug. it returns 1/1000 of actual value.
 	return time.Duration(bpMaxTimeLimitMs) * 1000
```
kroggen, line 140: I suggest to make test with the fixed value (`* 1000000`)

kroggen:
Why the forced push?
This makes the review very hard

Note also that are open conversations above that need to be processed for this PR to be merged

And it also needs to pass tests from #265 but implemented for `dpos` and `raft`

hayarobi:
> Why the forced push? This makes the review very hard
> 
> Note also that are open conversations above that need to be processed for this PR to be merged
> 
> And it also needs to pass tests from #265 but implemented for `dpos` and `raft`

I realized that my previous commit was wrong, so I immediately corrected the wrong code and pushed it again. In principle, it’s correct to make a new correct commit or a revert commit, but I replaced that commit for several reasons.
That commit was simple, didn't affect other branches or tags, and the two commits were so short apart (about 30 minutes or less) that no one had seen it yet. And since it is related to the merge commits, the readability of the commit graph would be worse if I proceeded in that order.

contract/vm.go
kroggen: is it useful to keep this time measurement?
maybe it can be commented to avoid consuming resources

consensus/impl/raftv2/blockfactory.go
```diff
@@ -122,7 +124,7 @@ type BlockFactory struct {
 
 	workerQueue chan *Work
 	jobQueue    chan interface{}
-	bpTimeoutC  chan struct{}
+	bpTimeoutC  chan struct{} // FIXME change to context based logic like dpos factory
```
kroggen, line 127: regarding this `FIXME`, was not `raft` already modified also?

consensus/impl/sbp/sbp.go
```diff
@@ -34,7 +35,8 @@ type txExec struct {
 func newTxExec(cdb consensus.ChainDB, bi *types.BlockHeaderInfo) chain.TxOp {
 	// Block hash not determined yet
 	return &txExec{
-		execTx: bc.NewTxExecutor(nil, contract.ChainAccessor(cdb), bi, contract.BlockFactory),
+		// FIXME
+		execTx: bc.NewTxExecutor(context.Background(), nil, contract.ChainAccessor(cdb), bi, contract.BlockFactory),
```
kroggen, line 39: Add some description of what to do to this FIXME

contract/contract.go
```diff
@@ -323,10 +315,8 @@ func preloadWorker() {
 		}
 
 		// create a new context
-		ctx := NewVmContext(bs, nil, nil, receiver, contractState, txBody.GetAccount(),
-			tx.GetHash(), request.bi, "", false, false, receiver.RP(),
-			request.preloadService, txBody.GetAmountBigInt(), txBody.GetGasLimit(),
-			txBody.Type == types.TxType_FEEDELEGATION)
+		// FIXME need valid context
+		ctx := NewVmContext(context.Background(), bs, nil, nil, receiver, contractState, txBody.GetAccount(), tx.GetHash(), request.bi, "", false, false, receiver.RP(), request.preloadService, txBody.GetAmountBigInt(), txBody.GetGasLimit(), txBody.Type == types.TxType_FEEDELEGATION)
```
kroggen, line 319: Can it be merged without the "valid context"?
If yes, maybe the FIXME can be removed?

contract/vm_dummy/vm_dummy_test.go
```diff
@@ -976,6 +976,8 @@ func xestInfiniteLoop(t *testing.T) {
 }
 
 func TestInfiniteLoopOnPubNet(t *testing.T) {
+	// FIXME delete skip after gas limit patch
+	//t.Skip("disabled until gas limit check is added")
```
kroggen, line 980: the `t.Skip` code is already commented, so maybe these 2 lines can be removed

kroggen:
I fixed a conflict on this branch and added some more review notes

consensus/impl/sbp/sbp.go
kroggen: So the fact that the integration tests are passing may be because there is no timeout check in `SBP` mode?
This only reinforces the need to run those tests on `dpos`
And/or maybe enable the above for `SBP`

hayarobi:
@kroggen There are still a few things left to improve this PR. However, in my opinion, it is better to merge this PR first, release 2.4.8, and patch them that need to be improved in the later release.
This PR is a patch that fixes a potentially dangerous bug that may still occur, and I believe that any remaining issues are not immediately urgent or critical. I will create cards for the remaining issues to issues pages.

kroggen (approved):

PR #263, opened 2023-09-25
pop module from stack after luaL_register

kroggen:
It was lacking to add `lua_pop()` on 2 places

### DIFF
--- a/contract/db_module.c
+++ b/contract/db_module.c
@@ -600,36 +600,36 @@ int lua_db_release_resource(lua_State *L) {
 	return 0;
 }
 
+static const luaL_Reg rs_methods[] = {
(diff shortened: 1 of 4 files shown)


hayarobi (approved):
lgtm

contract/db_module.c
hayarobi, line 1: It changes init positon of static variables. I'm too past to follow C code well. What is different?

kroggen: On all other modules these static variables are outside the function that use them, so I made this file have the same structure

kroggen: [here](https://github.com/aergoio/aergo/blob/develop/contract/system_module.c#L459) is an example

PR #265, opened 2023-09-27
some new tests

kroggen:
The integration test here is useful to test the block timeout check (#262) while contracts are deployed by other contracts


kroggen:
Lets merge #267 before this one

Then this PR can be updated to use the functions there

gokch (approved):

PR #266, opened 2023-09-27
deploy contracts as plain code

kroggen:

PR #267, opened 2023-09-27
new functions for integration tests

kroggen:
`aergocli` will work differently if deploying for <= V3 or >= V4

These new functions prepare the integration tests for new deploy method on #266

Although this is focused on changes to be applied only on V4, the changes here are backwards compatible and it is useful to merge them before so other work made on integration tests can use these new functions


hayarobi (approved):
lgtm

PR #269, opened 2023-09-28
bignum.ispositive() + bignum.isnegative()

kroggen:
This PR adds 2 new functions:

* `bignum.ispositive()`
* `bignum.isnegative()`

The last one is a clone of the `bignum.isneg()`. Both will exist (for backwards compatibility) but the docs and code completion on the web editor will mention only `bignum.isnegative()`


kroggen:
TODO: the tests must be run for all the hardfork versions

kroggen:
PROBLEM: it is giving difference in gas on V4 when new functions are added to the module

If we comment the lines that add the new functions, it gives the expected value. When we uncomment the lines (and the functions are added) it gives another amount of gas

The result can be seen [here](https://github.com/aergoio/aergo/actions/runs/6546556040/job/17777251175#step:7:1805)

hayarobi (approved):
lgtm

PR #270, opened 2023-09-28
add system.version()

kroggen:
This function returns the hardfork version of the underlying blockchain


kroggen:
TODO: add tests

kroggen:
It is intended to be used like this:

```lua
  if system.version() >= 4 then
    ...
  end
```

kroggen:
Another option to check if a function is available:

```lua
if type(bignum.ispositive) == "function" then
  ...
end
```

But it only works for functions on existing modules. This will fail if the `name_service` module is not available:

```lua
if type(name_service.resolve) == "function" then
  ...
end
```

> error: attempt to index global 'name_service' (a nil value)

This method checks for the module and for the function:

```lua
if name_service ~= nil and type(name_service.resolve) == "function" then
  ...
end
```

hayarobi (approved):
lgtm

PR #276, opened 2023-10-04
small refactoring about contract state

gokch:
# What is changed?
OpenContractStateAccount(XXX) to GetXXXAccountState()

# Is there risk of fork?
not at all.


kroggen:
Avoid renaming the title of editing the first message because these both trigger a new full test

The trigger should happen only when the base branch is changed, but unhappily it also occurs for the cases above

kroggen (approved):
Good

PR #287, opened 2023-10-12
Topic/schema

gokch:
## What is changed?
collect key prefix and key creation functions when access to db
make unit tests for key creation functions.

## Is there risk of fork?
yes. there are no changes, but need to sync test.


chain/chaindb.go
```diff
@@ -552,7 +531,7 @@ func (cdb *ChainDB) dropBlock(dropNo types.BlockNo) error {
 	dbTx.Delete(dropIdx)
 
 	// update latest
-	dbTx.Set(latestKey, newLatestIdx)
+	dbTx.Set([]byte(schema.Latest), newLatestIdx)
```
kroggen, line 534: The use of `[]byte(` on every `Get` and `Set` is annoying
The keys could be stored already in `[]byte` format

types/schema/schema.go
```diff
@@ -0,0 +1,49 @@
...
+	Latest         = ChainDBName + ".latest"
+	HardFork       = "hardfork"
+	ReOrg          = "_reorg_marker_"
+
+	// dpos
+	DposLibStatus = "dpos.LibStatus" // LibStatusKey is the key when a LIB information is put into the chain DB.
+
+	// raft
+	RaftPrefix             = "r_"
+	RaftIdentity           = RaftPrefix + "identity"
```
kroggen, line 23: The `RaftPrefix` can be left as `string`. The keys can be `[]byte`

types/schema/key.go
```diff
@@ -0,0 +1,65 @@
...
+
+func KeyVoteSort(key []byte) []byte {
+	return append([]byte(SystemVoteSort), key...)
+}
+
+func KeyVoteTotal(key []byte) []byte {
+	return append([]byte(SystemVoteTotal), key...)
+}
+
+func KeyVpr(i uint8) []byte {
```
kroggen, line 63: All functions names on this file should **end** with `Key` instead of starting with it

types/schema/key.go
```diff
@@ -0,0 +1,65 @@
...
+	// upper double check
+	return append([]byte(EnterpriseConf), bytes.ToUpper(conf)...)
+}
+
+func KeyName(name []byte) []byte {
+	// lower double check
+	return append([]byte(Name), bytes.ToLower(name)...)
+}
+
+func KeyParam(id []byte) []byte {
```
kroggen, line 42: The argument must be `string`

kroggen:
I like the idea of using `dbkey` instead of `schema`

Here is a comparison:

```go
data, err := scs.GetData(dbkey.Staking(who))
```

or

```go
data, err := scs.GetData(schema.StakingKey(who))
```

It makes it clear that is is accessing a database

contract/name/name.go
```diff
@@ -49,7 +48,7 @@ func UpdateName(bs *state.BlockState, scs *state.ContractState, tx *types.TxBody
 	if err != nil {
 		return types.ErrTxInvalidRecipient
 	}
-	creator, err := contract.GetData([]byte("Creator"))
+	creator, err := contract.GetData([]byte(schema.CreatorMeta))
```
kroggen, line 51: Without the `[]byte()`

contract/name/name.go
```diff
@@ -168,9 +165,7 @@ func registerOwner(scs *state.ContractState, name, owner, destination []byte) er
 }
 
 func setNameMap(scs *state.ContractState, name []byte, n *NameMap) error {
-	lowerCaseName := strings.ToLower(string(name))
-	key := append(prefix, lowerCaseName...)
-	return scs.SetData(key, serializeNameMap(n))
+	return scs.SetData(schema.KeyName(name), serializeNameMap(n))
```
kroggen, line 168: Using only `Name` is not clear enough. What about `AccountName`?
```go
return scs.SetData(dbkey.AccountName(name), serializeNameMap(n))
```

contract/system/param.go
```diff
@@ -41,15 +41,11 @@ func InitSystemParams(g dataGetter, bpCount int) {
...
-func genParamKey(id string) []byte {
-	return []byte("param\\" + strings.ToUpper(id))
-}
-
 func loadParam(g dataGetter) parameters {
 	ret := map[string]*big.Int{}
 	for i := sysParamIndex(0); i < sysParamMax; i++ {
 		id := i.ID()
-		data, err := g.GetData(genParamKey(id))
+		data, err := g.GetData(schema.KeyParam([]byte(id)))
```
kroggen, line 48: What about `SystemParam`?
```go
data, err := g.GetData(dbkey.SystemParam(id))
```
Changes on this file may cause conflict with other open PR

contract/system/staking.go
```diff
@@ -135,13 +133,11 @@ func (c *unstakeCmd) run() (*types.Event, error) {
 }
 
 func setStaking(scs *state.ContractState, who []byte, staking *types.Staking) error {
-	key := append(stakingKey, who...)
-	return scs.SetData(key, serializeStaking(staking))
+	return scs.SetData(schema.KeyStaking(who), serializeStaking(staking))
 }
 
 func getStaking(scs *state.ContractState, who []byte) (*types.Staking, error) {
```
kroggen, line 139: What about renaming `who` to `account` or `address`? Also on the function above

chain/chaindb.go
gokch: it is because []byte can`t be declared by const. if it declared just var, it can changed on other packages.

types/schema/key.go
gokch: ok.

chain/chaindb.go
gokch: I think declare string literal to const is safer than than declaring all keys as var.

contract/name/name.go
gokch: it is not only about account name, and also include "aergo.name".
it is key function for Name service in governance, so AccountName is Inappropriate.

gokch: maybe, GovernanceNameKey can be more clear. ( but it should change all governance key like enterprise, system, vote, staking )

contract/system/param.go
gokch: ok. if it changed, StakingKey / VoteKey / VoteSortKey, VprKey also changed to SystemXXXKey.

chain/chaindb.go
kroggen: Using `[]byte()` everywhere is awful. It was not like that before. It may have a way to fix it

kroggen: All these keys were `var` already, so no reason to make the usage more ugly.
Option 1: Continue using `var` for the keys
Option 2: Change the `Set` and `SetData` to accept string for the key

contract/enterprise/config.go
```diff
@@ -134,7 +133,7 @@ func setConfValues(scs *state.ContractState, key []byte, values []string) (*Conf
 }
 
 func setConf(scs *state.ContractState, key []byte, conf *Conf) error {
-	return scs.SetData(append(confPrefix, genKey(key)...), serializeConf(conf))
+	return scs.SetData(schema.KeyEnterpriseConf(genKey(key)), serializeConf(conf))
```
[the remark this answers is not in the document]
kroggen: Why was this marked as resolved? The `genKey()` function is still there...

contract/system/param.go
kroggen: It is way better to remove the `Key` from the names and use `dbkey` instead of `schema`
The above would be `dbkey.SystemParam(id)`
If a collision of variable or funcion name happens, just use another name for the them

chain/chaindb.go
kroggen: I like the idea of using functions to return the key:
* Solves the problem of collision of key name and function name
* Makes the code shorter without the need of `[]byte()`
* Creates a "standard": use function for keys both when there is and there is no argument

The keys can be stored as const string and with the word `Key` at the end of variable name and the functions without `Key` because there is already on the `dbkey` package

Example:
```go
const {
  ...
  LatestBlockKey = ChainDBName + ".latest"
  ...
}

...

func LatestBlock() []byte {
  return []byte(LatestBlockKey)
}
```

kroggen: And if the keys will not be accessed directly, they can be declared as private
Example: `latestBlockKey`

gokch: good idea. i`ll change as you said.

kroggen:
Perfect! 👏 

The only thing missing is to remove the `genKey()` function from `contract/enterprise/config.go`. The `EnterpriseConf()` function is already using `bytes.ToUpper()` so no need to make new changes.

kroggen (approved):

gokch:
> Perfect! 👏
> 
> The only thing missing is to remove the `genKey()` function from `contract/enterprise/config.go`. The `EnterpriseConf()` function is already using `bytes.ToUpper()` so no need to modify it.

good. but it need to sync test before merge. 

but genKey() is used about validate too. it seems good to remove genKey() in GetData() and SetData().

gokch:
after merge, i will make additional tests about change db key ( [fix/state-trie-keys](https://github.com/aergoio/aergo/tree/fix/state-trie-keys) branch ), and prepare to db migration.

contract/enterprise/validate.go
```diff
@@ -75,7 +75,7 @@ func ValidateEnterpriseTx(tx *types.TxBody, sender *state.V,
 		if err := checkArgs(context, &ci); err != nil {
 			return nil, err
 		}
-		key := genKey([]byte(context.Args[0]))
+		key := []byte(context.Args[0])
```
kroggen, line 78: I did not know that the function was used in other places. Above it is not converting to uppercase. Maybe the function can be bring back

kroggen:
I resolved conflicts after the merge of #259

contract/enterprise/validate.go
gokch: i check all key converting to uppercase in validate and execute. It's okay to revert, but it would be better to create unit tests for this.

kroggen:
@kslee8282 Please run sync test on this as well, if not done yet

kslee8282:
> @kslee8282 Please run sync test on this as well, if not done yet

2.4.9-rc3 end sync

PR #290, opened 2023-10-16
integration tests for dpos and raft

kroggen:
This does not require sync test, only test files are changed


kroggen:
Note: currently the test for `raft` is marked as public to be able to run the same gas tests used on other consensus, but it can be changed later to private. On this case all checks of used gas on all tests should be disabled if `consensus == "raft"`

kroggen:
@hayarobi Execution timeout is not happening when deploying 6 contracts on the same transaction on `dpos` and `raft`, as can be seen [here](https://github.com/aergoio/aergo/actions/runs/6536301701/job/17747693434#step:10:353)

gokch (approved):

hayarobi (approved):
lgtm

PR #291, opened 2023-10-17
fix wrong block generation inverval in raft

hayarobi:
It fix bug that the blocks are generated too much in raft consensus. 
It preserve behavior of v2.4.7 at best effort.


gokch:
it is hotfix pr including in v2.4.8.

gokch (approved):

PR #296, opened 2023-10-23
add vm_direct

kroggen:
This file is used by [tx-replay](https://github.com/aergoio/tx-replay) to execute transactions directly on the VM without using the chain, consensus and mempool packages


kroggen:
`vm_direct` may not be the best name. I just did not found another one for now
If you have any suggestion, feel comfortable to share it

kroggen:
I suspect that this new file on a separate folder does not affect the Aergo, as it is not used by it.

So this PR can be merged right away

gokch (approved):

PR #297, opened 2023-10-24
enhance the parsing of amount

kroggen:
This replaces the `transformAmount()` function with another one, which:

* Is easier to understand
* Do not allow repeated units (eg "100 aergo 20 aergo")
* Do not allow units in any order, only `aergo` then `gaer` then `aer`

The new function emits backwards compatible error messages

And also tests for the function


gokch:
> @hayarobi @gokch Please check what do you think about the changes on this PR, before running a sync test I do not know if there is a reason to allow repeated units or units in any order on amount strings

Great! I think repeated units don`t need too. if no problem in sync tests, it can changed.

gokch:
by the way, transformAmount can used in other packages too.
for example, aergocli also use [their own transformAmount](https://github.com/aergoio/aergo/blob/develop/cmd/aergocli/util/unit.go). it is duplicate logic.
it can moved to types/token_unit.go, and used by all place it needed. what do you think about that?

kroggen:
The aergocli ParseUnit supports decimal amounts, like "1.23 aergo"
I will update #177 once this one is merged. It adds support for decimal amounts on contracts as well
So maybe later they can use the same logic, on V4

kroggen:
@kslee8282 
This branch is ready for a sync test

kslee8282:
> @kslee8282 This branch is ready for a sync test

Oct 29 05:17:22.315 INF ../go/aergo/cmd/aergosvr/aergosvr.go:100 > AERGO SVR STARTED branch=topic/parse-amount module=asvr revision=dd63546a

testnet sync (26063000 / 111674681)
mainnet sync (122440984 / 142658661)

kslee8282:
sync success

gokch (approved):

PR #299, opened 2023-10-25
unit tests for contract package

kroggen:


kroggen:
@gokch 
I moved your code here to the `develop` branch and made modifications
On this way it can be used in different PRs later, after merging this one
I had to rename the `validateTxType` function to better reflect its usage. The code does not return `valid` but if the parent function should continue execution or not

gokch (approved):

PR #303, opened 2023-10-29
increase max events per txn

### LINKED ISSUE #302: Increase limit of events per transaction
The current limit is 50 events / txn

When the MaxCallDepth was increased to 64 this other limit was not changed

It may be interesting to make the limit of events to be somewhat like twice the MaxCallDepth, allowing for 2 events for every contract call

On this case, if approved, the new limit would be `128` events per transaction (on V4)

Is it OK? Is there any implication of this change?

kroggen:
Closes #302


hayarobi (approved):
The code looks good. @ultrafellen The remains is your decision on the policy.

PR #306, opened 2023-10-31
disable some lua functions

kroggen:
This PR removes some Lua functions, starting on hardfork 4:

* `package` module
* `require`
* `getfenv`
* `setfenv`
* `getmetatable`
* `setmetatable`
* `rawget`
* `rawset`
* `rawequal`
* `string.dump`

They continue existing for previous hardforks for compatibility with gas computations. The first 4 are also available when Aergo is build in debug mode

This PR also does not build some disabled packages (os, io, jit, debug) and so the code cannot be accessed in any way

It includes tests to make sure the disabled modules and functions are not available


kslee8282:
sync success

kslee8282:
Nov  1 05:18:42.040 INF ../go/aergo/cmd/aergosvr/aergosvr.go:100 > AERGO SVR STARTED branch=topic/disable-lua-require module=asvr revision=b9d3bb50

kroggen:
@kslee8282 
Are you sure the sync test was for this PR?
This one is only for V4

contract/vm_dummy/test_files/feature_isolation.lua
```diff
@@ -22,8 +22,10 @@ function override_functions()
     -- override contract.balance
     contract.balance = function() return "123" end
 
-    -- override the __add metamethod on bignum module
-    getmetatable(bignum.number(0)).__add = function(x,y) return x-y end
+    if getmetatable ~= nil then
+      -- override the __add metamethod on bignum module
+      getmetatable(bignum.number(0)).__add = function(x,y) return x-y end
```
hayarobi, line 27: I don't review deeper code about getmemtable, but it doesn't look like addtion rather substraction. Anyway, if this code did not cause any problem, we can leave this as is, since this commit don't change the code.

hayarobi (approved):
lgtm

contract/vm_dummy/test_files/feature_isolation.lua
kroggen: yes, this test overrides the add (`+`) operator with a subtraction (`-`) to check if that feature of overriding functions is possible. the feature is disabled by this PR

PR #309, opened 2023-11-03
refactoring for encoding logic

gokch:
## What is changed?
small refactoring about encode ( hex, base58, base58check, base64, gob )
replace encode logic to using internal package


kroggen:
Why is the `GobDecode` in the `common` package instead of together with the others in `enc`?

Is there some requirement for that?

kroggen:
The conversion from `enc.ToString` to `enc.B58Encode` makes it clear what is the encoding format.

I am just not sure if the long name would not be better, like `Base58` instead of `B58`

---

I would prefer to use names like this instead of `enc`:

```
base64.Encode
base58.Encode
base58check.Encode
hex.Encode
```

So all would have a consistent usage, with {format}.(Encode|Decode)

This would require to have each of these inside of "enc", like "enc/base64", "enc/hex"...

kroggen:
In [types/common.go](https://github.com/aergoio/aergo/blob/14d88329b0102f7b5c7888779357ae02e048e98b/types/common.go) we can still find `EncodeB64` and `EncodeB58`

gokch:
> Why is the `GobDecode` in the `common` package instead of together with the others in `enc`?
> 
> Is there some requirement for that?

in before enc package was just b58 encode / decode. i`ll move gob function from common to enc.

gokch:
> In [types/common.go](https://github.com/aergoio/aergo/blob/14d88329b0102f7b5c7888779357ae02e048e98b/types/common.go) we can still find `EncodeB64` and `EncodeB58`

internal package doesn`t reveal for other repository, but indexer / tx_replay should use encode/decode logic. so it is good to leave in types.

kroggen (approved):

kroggen:
It appears OK to me

@hayarobi  What do you think?

gokch:
it need sync test before merged @kslee8282

kslee8282:
> it need sync test @kslee8282

Nov  6 06:07:57.931 INF ../go/aergo/cmd/aergosvr/aergosvr.go:100 > AERGO SVR STARTED branch=feature/encoding module=asvr revision=04c5a261
start sync mainnet

kroggen (approved):

kslee8282:
sync success

kroggen:
Simple conflict resolved

PR #312, opened 2023-11-09
disable CI run on PR edit

kroggen:
When the title or the first comment of a PR is edited, a new CI/CD workflow run is activated

This is annoying, uses CI time without need, and marks some PRs as failing when the run is cancelled

This PR disable this behavior

When the base of a PR is modified, then we need to start a new run manually, until GitHub has a solution for the problem


gokch:
What do you think about remove verbose option when go tests? Verbose test is confusing and make hard to find error line.

kroggen:
I agree

kroggen:
@hayarobi 
Do you agree?
Should the verboseness be removed also from the "code coverage" test?

hayarobi (approved):
lgtm

PR #318, opened 2023-11-16
Fix occasional failure of unit test

hayarobi:
This test was sometimes failed by timing issue, such like in heavy load in host machine of container.

### DIFF
--- a/consensus/impl/dpos/blockfactory_test.go
+++ b/consensus/impl/dpos/blockfactory_test.go
@@ -21,11 +21,11 @@ func TestBlockFactory_context(t *testing.T) {
 	select {
 	case <-timeoutCtx.Done():
 		err := timeoutCtx.Err()
-		assert.Equal(t, context.DeadlineExceeded, err, "err by deadline")
+		assert.Equal(t, context.DeadlineExceeded, err, "unexpected error type")
 		// check if timeout occured not before deadline
 		assert.True(t, time.Now().After(contextDeadline))
-	case <-time.NewTimer(time.Second * 5).C:
-		assert.Fail(t, "deadline expected, but not")
+	case <-time.NewTimer(5 * time.Second).C:
+		assert.Fail(t, "timeout did not occur within expected time frame")
 	}
 
 	// second channel is canceled by self cancel func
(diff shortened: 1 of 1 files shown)


consensus/impl/dpos/blockfactory_test.go
```diff
@@ -13,17 +13,18 @@ func TestBlockFactory_context(t *testing.T) {
...
 	// first channel is done by deadline
 	select {
 	case <-timeoutCtx.Done():
 		err := timeoutCtx.Err()
 		assert.Equal(t, context.DeadlineExceeded, err, "err by deadline")
-		//assert.Equal(t, err, context.Cause(timeoutCtx), "cause and err is differ")
-	default:
+		// check if timeout occured not before deadline
+		assert.True(t, time.Now().After(contextDeadline))
+	case <-time.NewTimer(time.Second * 5).C:
```
kroggen, line 27: replace to `5 * time.Second` to make it consistent with other places

consensus/impl/dpos/blockfactory_test.go
```diff
@@ -13,17 +13,18 @@ func TestBlockFactory_context(t *testing.T) {
...
 		assert.Fail(t, "deadline expected, but not")
```
kroggen, line 28: consider completing this sentence. it is not clear

kroggen (changes requested):

consensus/impl/dpos/blockfactory_test.go
hayarobi: Like below? 
```Context should be done by deadline.```

kroggen: "context ... done"? I don't understand it

kroggen: "deadline expected, but not happen" may be fine

hayarobi: The channel is done if its lifecycle is done by timeout or called cancel. And, `timeoutCtx.Done()` will block until the timeoutCtx is done. 

I can't have more ideas to clearly reveal what makes this test in failure. I will appreciate if you give me a good phrase.

kroggen: The one I wrote above is not good?

hayarobi: My reply was on RACE CONDITION. 😆

kroggen (approved):

PR #325, opened 2023-11-26
tidy packages

gokch:
## What is Changed?
remove unnecessary files - dockerfile, travis 
move to tools - example, polaris
move to types - message

## Is there risk of fork?
no


kroggen:
There is code on the `travis.ci` that is not moved to another place, mainly the generation of release packages (see on the `before_deploy` section)

The last releases do not include binaries

At least we can access the file via this PR if we need it later

gokch:
> If `aergosvr` uses polaris client, then it is not an external tool, right?

i agree. it seems good to move in cmd, not tools.

kroggen (approved):
It appears OK to me

PR #333, opened 2023-12-06
Small improvement of automation

hayarobi:
- actions/setup-go will not work without quote of go version
- repository of indirectly dependent library was changed.


hayarobi:
Some libraries depend on go-diff and appdash. though my memory is vague, it looks like that a package of libp2p or multiaddr uses appdash, and appdash depends on go-diff.
The source repository of these two package was moved from their own to github, so I directed the new repository location by `replace` directive.

kroggen (approved):

PR #336, opened 2023-12-14
Change log message of unidentified http request for investigation

hayarobi:
As a guess, it seems that originally the prev code was assumed that only grpcweb was supported, so if an http request other than grpc came, it could be an incorrect request or an impure intention, so it recorded logs at INFO level.
However, many other requests are coming in, so there are quite a lot of logs recorded on the testnet nodes, but there is no context in that log messages, so a lot of meaningless messages are piling up.
This commit is not a solution of the problem. The code has been modified to leave more detailed logs for analysis, and a commit that fixes the real problem will be recreated using the logs produced by this commit.


hayarobi:
Finally, I, my teammate actually, found the cause of wrong http request. The proxy configuration of legacy version of aergoscan. 
Anyway, we could find it by these commits which added more detailed log messages. I think it would be better to apply this PR as it will help find the cause in similar cases in the future.

kroggen (approved):

PR #341, opened 2024-01-16
add state dump

gokch:
# What is changed?
support dump for account, contract, contract storage in state trie

# Is there risk of fork?
no. it is for debug and test only.


kroggen:
Is this for the API?
Or to be used locally?

gokch:
> Is this for the API? Or to be used locally?

it used only locally.
maybe it can used when error occured.

kroggen (approved):
Appears to not interfere with existing code

kroggen:
There is also this: [aergoio/state-tools](https://github.com/aergoio/state-tools)

gokch:
snapshot migration is in progress, so it can add after 2.5.0 release.

PR #200, opened 2023-05-09
fix: use prefix on state trie keys

kroggen:

PR #342, opened 2024-01-18
check legacy trie key

gokch:
# What is changed?
check legacy trie key format ( related with #200 )

# Is there risk of fork?
no.


kroggen (approved):
OK

PR #348, opened 2024-02-07
rollback receipt and event format

gokch:
# What is changed?
rollback some json format for cli / web3 ( receipt, event )

# Is there risk of fork?
no.


kroggen (approved):
Works great!

kroggen:
Ops! There is a problem with the tests

PR #353, opened 2024-04-12
rollback partial web3 logic

waltzforzizi:
remove user-write code in proto files
regenerate proto and grpc files


hayarobi (approved):
lgtm

kroggen (approved):
OK

PR #356, opened 2024-05-23
aergoluac: add feature of decode payload

kroggen:
Decode payload into bytecode and ABI

The payload can come from:

* `aergoluac --payload`
* `aergocli` JSON transaction
* AergoScan (Hex format)


kroggen:
This was added to the `hardfork-v4` branch

PR #357, opened 2024-06-01
add blacklist

kroggen:


shepelt (approved):
Changes seem minimal and non-destructive.
Should be suitable before next hard fork.

PR #359, opened 2024-06-13
build docker image from local folder

kroggen:
Instead of using `git clone` and retrieving from repository

It can be used to build when the local folder has changes that were not committed, for testing purposes


hayarobi (approved):
lgtm

PR #360, opened 2024-06-19
Improve mpdumpdiag

hayarobi:
- Add flag to change format of payload.


kroggen:
what is this tool used for?

I did not know it existed

should it be build together with other tools when running `make`? what about building it only when used `make mpdumpdiag`? to avoid making the build process longer

hayarobi:
> should it be build together with other tools when running `make`? what about building it only when used `make mpdumpdiag`? to avoid making the build process longer

If just `make` also build mpdumpdiag , that is my mistake.

kroggen (approved):

PR #361, opened 2024-07-07
save contract source code

kroggen:
Save the source code of the deployed contract (both via transaction and via `contract.deploy()`) on the contract state, starting on hardfork V4

The previous method for V4 was to just compile the source code from the transaction and store only the bytecode

Now both bytecode and source code would be stored

This can be useful in the future if we want to use a Lua interpreter instead of LuaJIT (for some contracts)


hayarobi (approved):
lgtm

PR #363, opened 2024-08-22
fix some function names

curlwget:


hayarobi:
Thanks for your PR. 
The name of function `GRPCSeverBuilder()` is definitely a typo. We will merge this PR after renaming that function.

PR #365, opened 2024-10-03
internal operations

kroggen:
* Records the internal operations in the node database
* Encode them as JSON string, all transactions from a block (store just once per block to reduce I/O, using the block number as key)
* Adds a RPC endpoint to retrieve these values

### DIFF
--- a/chain/chaindb.go
+++ b/chain/chaindb.go
@@ -691,12 +691,28 @@ func (cdb *ChainDB) GetChainTree() ([]byte, error) {
 	return jsonBytes, nil
 }
 
-func (cdb *ChainDB) writeReceipts(blockHash []byte, blockNo types.BlockNo, receipts *types.Receipts) {
+func (cdb *ChainDB) writeReceiptsAndOperations(block *types.Block, receipts *types.Receipts, internalOps string) {
+	hasReceipts := len(receipts.Get()) != 0
+	hasInternalOps := len(internalOps) != 0
+
+	if !hasReceipts && !hasInternalOps {
+		return
+	}
+
 	dbTx := cdb.store.NewTx()
 	defer dbTx.Discard()
 
-	val, _ := gob.Encode(receipts)
-	dbTx.Set(dbkey.Receipts(blockHash, blockNo), val)
+	blockHash := block.BlockHash()
+	blockNo := block.BlockNo()
+
+	if hasReceipts {
+		val, _ := gob.Encode(receipts)
+		dbTx.Set(dbkey.Receipts(blockHash, blockNo), val)
+	}
+
+	if hasInternalOps {
+		dbTx.Set(dbkey.InternalOps(blockNo), []byte(internalOps))
+	}
 
 	dbTx.Commit()
 }
--- a/chain/chaindb.go
+++ b/chain/chaindb.go
@@ -665,6 +665,11 @@ func (cdb *ChainDB) checkExistReceipts(blockHash []byte, blockNo types.BlockNo) 
 	return true
 }
 
+func (cdb *ChainDB) getInternalOperations(blockNo types.BlockNo) string {
+	data := cdb.store.Get(dbkey.InternalOps(blockNo))
+	return string(data)
+}
+
 type ChainTree struct {
 	Tree []ChainInfo
 }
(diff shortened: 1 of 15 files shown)


chain/chainhandle.go
```diff
@@ -1039,7 +1052,7 @@ func executeTx(execCtx context.Context, ccc consensus.ChainConsensusCluster, cdb
 			}
 			return types.ErrNotAllowedFeeDelegation
 		}
-		rv, events, txFee, err = contract.Execute(execCtx, bs, cdb, tx.GetTx(), sender, receiver, bi, executionMode, true)
+		rv, events, _, txFee, err = contract.Execute(execCtx, bs, cdb, tx.GetTx(), sender, receiver, bi, executionMode, true)
```
hayarobi, line 1055: feeDelegation executes the contract like a normal transaction execution, except that the contract wallet pays the transaction fee. In this case, is it not necessary to record internalOps? Or is there no possibility of internalOps occurring in a feeDelegation transaction?

chain/chaindb.go
```diff
@@ -691,18 +696,35 @@ func (cdb *ChainDB) GetChainTree() ([]byte, error) {
...
-	dbTx.Set(dbkey.Receipts(blockHash, blockNo), val)
+	blockHash := block.BlockHash()
+	blockNo := block.BlockNo()
+
+	if hasReceipts {
+		val, _ := gob.Encode(receipts)
+		dbTx.Set(dbkey.Receipts(blockHash, blockNo), val)
+	}
+
+	if hasInternalOps {
```
hayarobi, line 718: It looks like that adding new field into database. Does this PR needs to reconstruct stateDB?

kroggen: As I described above, it saves all internal operations as a JSON string into the database, one record for each block.

It does not affect the state db, it is stored on the chain database

The reason to store per block instead of per txn, is to decrease I/O

When the indexer sees that a block has transactions, it must send a request to the node to retrieve the list of internal operations

chain/chainhandle.go
kroggen: Good catch!

Transactions with fee delegation are executed in 2 steps:

1. (the code above) the mempool executes the contract (`check_delegation` function) to check if it will pay for the fees
2. if yes, then the txn is included in a block and then executed normally

Currently the code is logging the internal operations only from the normal execution

I also suspect that on step 1 it is not allowed to make some operations like call other contracts, deploy, etc. It can only read from the state variables.

chain/chaindb.go
kroggen: Oh yes, the nodes that will have this feature enabled need to be sync from block 1 again, because it needs to re-execute all the txns to extract the internal operations

I suspect that this feature will be enabled only on RPC watcher nodes, not on GP nodes. But this needs to be decided

hayarobi:
All tests need to be passed.

kroggen:
It is passing now.
There is no tests specific to this feature, though.

Note that this is not intended to be merged now. It lacks a sync test, and manual tests with the indexer.
It can be merged into `develop` later

hayarobi (approved):
lgtm

hayarobi (approved):
Just consider naming of config field.
The others are good to me.

config/types.go
```diff
@@ -53,6 +53,8 @@ type RPCConfig struct {
 	NSKey       string `mapstructure:"nskey" description:"Private Key file for RPC or REST API"`
 	NSCACert    string `mapstructure:"nscacert" description:"CA Certificate file for RPC or REST API"`
 	NSAllowCORS bool   `mapstructure:"nsallowcors" description:"Allow CORS to RPC or REST API"`
+	// Internal operations
+	LogInternalOperations bool `mapstructure:"internal_operations" description:"Log internal operations"`
```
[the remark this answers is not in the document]
kroggen: No problem for me
But the config file of existing nodes would need to be updated
I guess they are few anyway

hayarobi (approved):
It was already approved and just merged another approved PR.

config/types.go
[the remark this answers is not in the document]
kroggen: Feel free to rename the config field. I am working on other features now

ultrafellen (approved):

PR #368, opened 2024-10-21
use proper error message for deploy of bytecode

kroggen:


kroggen:
I suspect that this can be reopened. We can make a sync test with this

hayarobi (commented):
I think it's better to use the WARNING level in an unexpected situation or a strong suspicion of abnormalities. The failure of deploying contract is mostly caused that the user outside the system sends the wrong data. And, the server node filter it in a normal verification process, so it will be more appropriate to lower the level to INFO.

kroggen:
OK, updated

hayarobi (approved):
lgtm

PR #370, opened 2024-11-05
Added Hardfork history field to ChainInfo rpc.

hayarobi:
- The keys and values are the hardfork version and the block height at which the hardfork starts.

I will change the commit hash of submodule aergo-protobuf after merging PR https://github.com/aergoio/aergo-protobuf/pull/42 in aergo-protobuf.


chain/stubchain.go
```diff
@@ -218,6 +218,11 @@ func (tchain *StubBlockChain) ChainID(bno types.BlockNo) *types.ChainID {
 	return nil
 }
 
+func (tchain *StubBlockChain) HardforkHeights() map[string]types.BlockNo {
+	// FIXME: maybe it should return at least latest hardfork
+	return nil
```
kroggen, line 223: So this PR is not ready?

chain/chaindb.go
```diff
@@ -757,12 +758,42 @@ func (cdb *ChainDB) Hardfork(hConfig config.HardforkConfig) config.HardforkDbCon
...
+		logger.Error().Msg("Failed to read hardfork from cdb")
+		return c
+	}
+	returned := make(map[string]uint64, len(c))
+	for key, value := range c {
+		newKey := key[1:]
+		if key[0] != 'V' || !IsInteger(newKey) {
+			return make(map[string]uint64, 0)
+		}
+		returned[newKey] = value
```
kroggen, line 785: I like that you are using just the version number without the "V"
But there are 2 parts in the code bellow that have the V. Is that OK?

chain/chaindb_test.go
```diff
@@ -0,0 +1,97 @@
...
+		store     db.DB
+	}
+	tests := []struct {
+		name   string
+		hardfork string
+		wantSuccess bool
+	}{
+		{ "normal", `{"V2":2000,"V3":30000,"V4":400000}`, true},
+		{ "wrong", `{"A":2000,"B3A":30000,"V4":400000}`, false},
+		{ "wrong2", `{"A2":2000,"B3":30000,"V4":400000}`, false},
```
kroggen, line 25: here is one place

types/jsonrpc/chaininfo_test.go
```diff
@@ -20,6 +20,7 @@ func TestConvChainInfo(t *testing.T) {
 		namePrice        = types.NewAmount(1, types.Aergo)
 		totalVotingPower = types.NewAmount(10000000, types.Aergo)
 		votingReward     = types.NewAmount(1, types.Aergo)
+		hardfork         = map[string]uint64{"V1": 100000, "V2": 200000, "V3": 3000000}
```
kroggen, line 23: and here is another

kroggen (approved):

chain/chaindb_test.go
hayarobi: This is simulate the chaindb, not rpc response. 
The chaindb stores hardfork status with json string since hardfork V2. We cannot change the internal data structures without migration.

types/jsonrpc/chaininfo_test.go
hayarobi: See reply above.

PR #372, opened 2024-11-12
Use a more direct and less error-prone return value

pinglanlu:
Since err != nil has been judged before, nil is returned directly here, which is more obvious, readable and less error-prone.


kroggen (approved):
OK

pinglanlu:
> OK

Thank you for your approval.

PR #373, opened 2024-11-26
fix integration tests

kroggen:
fix sporadic failure by making all RPC calls be sent to the same node

### DIFF
--- a/tests/common.sh
+++ b/tests/common.sh
@@ -95,7 +95,7 @@ get_receipt() {
   set +e
 
   while true; do
-    output=$(../bin/aergocli receipt get --port $query_port $txhash 2>&1 > receipt.json)
+    output=$(../bin/aergocli receipt get $txhash 2>&1 > receipt.json)
 
     #echo "output: $output"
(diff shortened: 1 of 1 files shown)


kroggen:
@hayarobi 
Can you review this?

tests/common.sh
```diff
@@ -95,7 +95,7 @@ get_receipt() {
   set +e
 
   while true; do
-    output=$(../bin/aergocli receipt get --port $query_port $txhash 2>&1 > receipt.json)
+    output=$(../bin/aergocli receipt get $txhash 2>&1 > receipt.json)
```
hayarobi, line 98: Does this guarantee connection to the same node? I suspect that some real change is missing in this PR.

kroggen: by removing this, it will connect to the default port
this is what happens on other parts of the integration test
so all queries will all be made to the same port, and thus to the same node (because each node uses a different port)

hayarobi (approved):
lgtm

PR #375, opened 2024-12-26
fix compiled protobuf for SourceHash field

kroggen:
This PR fixes a bug in which a protobuf file was edited manually instead of being recompiled

It was done using the same previous version of `protoc` and `protoc-gen-go` to generate minimal code changes


hayarobi (approved):
lgtm

PR #376, opened 2024-12-27
add hardfork history field to ChainInfo RPC - 2

kroggen:
This is a modified version of commit abbedddba93c8301001e7eb988590d73738662c4 but with just the required changes to the protobuf file, and using the same version as before for easier comparison

It also does not include changes for Internal Operations, that is independent from this PR

And does not include the compilation of the SourceHash field, that should be done in the `develop` branch - check #375


hayarobi (approved):
lgtm

PR #383, opened 2025-02-20
fix warning on dockerfiles

kroggen:
Fix the warning:
```
FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 1)
```


kroggen:
I can't merge this without your review
It is almost a no-op change, just modifying the letters to uppercase

hayarobi (approved):
lgtm

PR #384, opened 2025-03-06
check if contract is blacklisted on deploy with address

kroggen:


contract/vm_callback.go
```diff
@@ -1142,6 +1143,11 @@ func luaDeployContract(
 		if err != nil {
 			return -1, C.CString("[Contract.LuaDeployContract]" + err.Error())
 		}
+		// check if contract is blacklisted
+		if blacklist.Check(contractStr) {
```
hayarobi, line 1147: I'm not sure whether the variable `contractStr` is address of contract. And, I think it's ok if that variable is string of contract address.

hayarobi (approved):
See my comment and then merge if you confirm.

contract/vm_callback.go
kroggen: the `contractStr` contains the first argument to `contract.deploy(source_or_address, arg1, arg2...)`

it can be either an address of an existing contract or the source code for a new contract

the `blacklist.Check()` will only return `true` in the case that it is an address AND it is included on the blacklist

PR #391, opened 2025-03-23
fix error message on transfer to smart contract

### LINKED ISSUE #143: Better error message on transfer to smart contract
When making a simple transfer to a smart contract by using a transaction of type `TRANSFER`:

The transfer fails (as expected) if the contract does not have a `payable` and `default` function.

But the error message is strange as the engine treats the transaction of type `TRANSFER`  as a transaction of type `CALL`:

https://testnet.aergoscan.io/transaction/Fu7eCFB6BnA7xKtuhYjtGkHfzBfZUah7p7iCeCpeqbXs?receipt=json&payload=formatted

```
"result": "not found function: "
```

### Proposal

If the transaction is of type `TRANSFER` then the engine:

* Could return a proper error message, like "the contract does not have a payable default function"
* Could ignore the payload (maybe)

kroggen:
the current error message when making a transfer (send) to a contract without a 'default' function is:

```
"not found function: "
```

this will change it to: (on the next hardfork)

```
"the contract does not have a payable default function"
```

Fixes #143

### DIFF
--- a/contract/vm.go
+++ b/contract/vm.go
@@ -282,6 +282,10 @@ func resolveFunction(contractState *statedb.ContractState, bs *state.BlockState,
 	if len(name) == 0 && defaultFunc != nil {
 		return defaultFunc, nil
 	}
+	// the function was not found
+	if currentForkVersion >= 5 && len(name) == 0 {
+		return nil, errors.New("the contract does not have a payable default function")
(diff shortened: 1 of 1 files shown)


hayarobi (approved):
lgtm

PR #392, opened 2025-03-27
deploy: add support for import statement on aergocli and brick

kroggen:
Contracts deployed with `brick` and `aergocli` now support the `import` statement in the following format:
```js
import 'file.lua'
import "path/to/another/file.lua"
import "https://raw.githubusercontent.com/aergoio/aergo-contract-ex/refs/heads/master/contracts/typecheck/typecheck.lua"

...
```
So there is no need to use `ship`

PR #393, opened 2025-03-28
process import statements - part 2

kroggen:
This is the second part of #392

It requires a sync test before merge because it modifies `vm.go`

It adds the support for the import statements on vm_dummy tests

It also separates code on the `aergoluac` module into 2 packages: `util` and `luac`

Only `luac` uses LuaJIT. Importing only the `util` will not add LuaJIT to the user project


cmd/aergoluac/util/util.go
hayarobi, line 1: Was luac_util.go file renamed to util.go ?

kroggen: the original `luac_util` was separated into 2 modules: luac and util

I put information on the description above

kroggen:
This requires a sync test, so it will not be merged now

hayarobi (approved):
lgtm

PR #394, opened 2025-03-28
aergocli: add limits to event stream

kroggen:
This PR adds 2 limits to event stream on `aergocli`:

* Limit of events to receive
* Limit of time to wait (timeout)


kroggen:
@hayarobi I also need your review there to merge this

It is tested and working

PR #395, opened 2025-03-31
fix p2p segfault

kroggen:
This is an attempt to fix a nil pointer dereference that is happening on tests lately

One case here:

https://github.com/aergoio/aergo/actions/runs/14141376620/job/39623139654

### DIFF
--- a/p2p/blkreceiver.go
+++ b/p2p/blkreceiver.go
@@ -159,7 +159,9 @@ func (br *BlocksChunkReceiver) cancelReceiving(err error, hasNext bool) {
 			case <-br.senderFinished:
 				break
 			}
-			br.peer.ConsumeRequest(br.requestID)
+			if br.peer != nil {
+				br.peer.ConsumeRequest(br.requestID)
+			}
 		}()
 	}
 }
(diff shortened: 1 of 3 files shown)


p2p/blkreceiver.go
```diff
@@ -159,7 +159,9 @@ func (br *BlocksChunkReceiver) cancelReceiving(err error, hasNext bool) {
 			case <-br.senderFinished:
 				break
 			}
-			br.peer.ConsumeRequest(br.requestID)
+			if br.peer != nil {
```
hayarobi, line 162: The br.peer itself if not null since it was written in constructor only and is not set anywhere. If the seg. fault occurs on test, it can be related to mocking framework, I guess.

p2p/hashreceiver.go
```diff
@@ -130,7 +130,9 @@ func (br *BlockHashesReceiver) cancelReceiving(err error, hasNext bool) {
 			case <-br.senderFinished:
 				break
 			}
-			br.peer.ConsumeRequest(br.requestID)
+			if br.peer != nil {
```
hayarobi, line 133: It is also immutable like blkreceiver.

p2p/blkreceiver.go
kroggen: Ah, OK. I can cancel this PR.

I just don't understand why it is failing just now, these sporadic failures did not appear in the past. Maybe it is related to the update of `libp2p`

Can you take a look on the mocking framework?

PR #396, opened 2025-04-02
attempt to fix force reset height

kroggen:
when `forceresetheight` is used, `aergosvr` will exit just after removing new blocks from the db


kroggen:
@hayarobi This one here does not require sync test as it is only used when truncating the local database

hayarobi (approved):
lgtm

PR #397, opened 2025-04-02
fix build in debug mode

kroggen:
make the packages available when built in debug mode

this is useful for `brick`, to run unit tests and functional tests

### DIFF
--- a/contract/vm.c
+++ b/contract/vm.c
@@ -57,6 +57,7 @@ static void preloadModules(lua_State *L) {
 		luaopen_db(L);
 	}
 
+#ifndef DEBUG
 	if (vm_is_hardfork(L, 4)) {
 		lua_getglobal(L, "_G");
 		// disable getmetatable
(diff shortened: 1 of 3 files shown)


contract/vm.c
```diff
@@ -57,6 +57,7 @@ static void preloadModules(lua_State *L) {
 		luaopen_db(L);
 	}
 
+#ifndef DEBUG
```
hayarobi, line 60: I'm confused that this is configuration for debug or for enabling/disabling a feature of hardfork v4

kroggen: it works in this way:

if compiled in `release` mode, the functions will be disabled only after hardfork 4. this is the current behavior

if compiled in `debug` mode, the functions will always be available, regardless of hardfork version

kroggen: this requires a sync test, it will not be merged now

kroggen: I am also trying to make `brick` to print the line numbers when an error occurs, and even printing the stack trace, but without any success so far

I modified to not strip the debug symbols when compiling the Lua code (only in debug mode) and added code to return stack trace, but it still does not work

hayarobi (approved):
good

hayarobi (approved):

PR #399, opened 2025-04-09
fix integration test with previous version

kroggen:
use `aergo/node:2.6.1` instead of `aergo/node:2.6.0` on the integration tests


hayarobi (approved):
good job

PR #404, opened 2025-04-24
Improve logging

hayarobi:
- Update aergo-lib and other libraries
- shorten length of file path by applying new feature of aergo-lib
- Tune some log levels
- Fix typo of log messages


cmd/aergosvr/aergosvr.go
```diff
@@ -24,6 +22,7 @@ import (
 	"github.com/aergoio/aergo/v2/rpc/web3"
 	"github.com/aergoio/aergo/v2/syncer"
 	"github.com/spf13/cobra"
+	"os"
```
kroggen, line 25: I suspect this line should be on the original place

consensus/impl/raftv2/p2p.go
```diff
@@ -19,7 +19,7 @@ var (
 	ErrGetClusterFail    = errors.New("failed to get cluster info")
 )
 
-// GetBestBlock returns the current best block from chainservice
+// GetClusterInfo sends collects cluster info from remote peers via p2p module
```
kroggen, line 22: I suspect that `sends` can be removed

p2p/raftsupport/concclusterreceiver.go
```diff
@@ -194,19 +196,19 @@ func (r *ConcurrentClusterInfoReceiver) calculate(err error) *message.GetCluster
 	if err != nil {
 		rsp.Err = err
 	} else if len(r.succResps) < r.requiredResp {
-		rsp.Err = errors.New("too low responses: " + strconv.Itoa(len(r.succResps)))
+		rsp.Err = errors.New("too low responses: " + strconv.Itoa(len(r.succResps)) + " , required " + strconv.Itoa(r.requiredResp))
```
kroggen, line 199: `low` or `few`?

kroggen: BTW this is not logging but returning a different error message than before. Can it cause hardfork when used with older nodes that return a different error message?

p2p/raftsupport/concclusterreceiver.go
```diff
@@ -19,7 +19,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-// ClusterInfoReceiver is send p2p getClusterInfo to connected peers and Receive p2p responses one of peers return successful response
+// ConcurrentClusterInfoReceiver is a class which send p2p getClusterInfo to connected peers and receive responses one of peers return successful response.
 // The first version will be simplified version. it send and Receive one by one.
```
kroggen, line 23: What about this?
```
// ConcurrentClusterInfoReceiver is a struct that manages concurrent requests for cluster information from peers.
// It sends p2p GetClusterInfo requests to connected peers and collects responses until either:
//   - A successful response is received
//   - The required number of responses are collected
//   - The operation times out
//
// The initial implementation processes requests and responses sequentially (one by one).
```

kroggen: Optionally, the struct below can also have comments for each field:
```go
type ConcurrentClusterInfoReceiver struct {
	logger *log.Logger      // Logger for recording events and errors
	mf     p2pcommon.MoFactory // Factory for creating message objects
	peers  []p2pcommon.RemotePeer // List of connected peers to query
	
	mutex   sync.Mutex      // Mutex to protect concurrent access to the struct
	sent    map[p2pcommon.MsgID]p2pcommon.RemotePeer // Tracks sent requests by message ID
	sentCnt int             // Count of total sent requests
	
	req    *message.GetCluster // The cluster information request message
	ttl    time.Duration    // Time-to-live duration for the operation
	timeout time.Time       // Absolute time when the operation will timeout
	
	respCnt      int        // Count of received responses
	requiredResp int        // Minimum required responses before completing
	succResps    map[types.PeerID]*types.GetClusterInfoResponse // Successful responses by peer ID
	
	status   receiverStatus // Current state of the receiver
	finished chan bool      // Channel to signal operation completion
}
```

consensus/impl/raftv2/p2p.go
hayarobi: It's a simple typo, I'll correct it.

p2p/raftsupport/concclusterreceiver.go
hayarobi: Your comments looks more readable than mine. I'll apply them.

hayarobi: My IDE, Goland, forcefully changes comments with same indent per section like below. I can find a way to solve it by consuming more time, but I think it's not as important as that.
```go
	logger *log.Logger            // Logger for recording events and errors
	mf     p2pcommon.MoFactory    // Factory for creating message objects
	peers  []p2pcommon.RemotePeer // List of connected peers to query

	mutex sync.Mutex // Mutex to protect concurrent access to the struct

	sent    map[p2pcommon.MsgID]p2pcommon.RemotePeer // Tracks sent requests by message ID
	sentCnt int                                      // Count of total sent requests

	req     *message.GetCluster // The cluster information request message
	ttl     time.Duration       // Time-to-live duration for the operation
	timeout time.Time           // Absolute time when the operation will timeout

	respCnt      int // Count of received responses
	requiredResp int // Minimum required responses before completing
	// Successful responses by peer ID
	succResps map[types.PeerID]*types.GetClusterInfoResponse

	status   receiverStatus // Current state of the receiver
	finished chan bool      // Channel to signal operation completion
```

p2p/raftsupport/concclusterreceiver.go
hayarobi: It's about count of responses, so few is more accurate as you said.
I will not cause hardfork because this message is not written in block, and don't affect executions of transactions.

kroggen (approved):
OK

PR #405, opened 2025-05-14
hardfork 5

kroggen:


hayarobi (approved):
lgtm.

kroggen:
@hayarobi You will need to review this one again

But before that, we need to add the block heights in which the hardfork should be activated

hayarobi:
The codes seems to be no problem except for the height of the hard fork. 

There are quite a lot of branches, including this PR, in a long time ago. If they are merged at the same time the commit graph will become very complex making hard to track feature chages. If the final codebase after the merge is the same, how about rebase each PR and squash the commits by a function unit and then merge? I will show the example soon.

hayarobi:
This is expectation of merging all PR by merge commit.
with-merge-orig

This is expectation of merging PR by rebasing and squashing small fix commits into one commit. 
with-rebase-orig

kroggen:
OK, it can be done via rebase and squash

hayarobi (approved):
Finally, changed the heights of hardfork 5 to the official heights.

PR #407, opened 2025-06-20
Refactoring dump account features

hayarobi:
- new dump function for flexible use.


kroggen (approved):

kroggen:
Would not be better to have this dump code on state-tools?

hayarobi:
This should be in aergosvr, since it uses private scoped member variables and methods of StateDB class. The state-tools can take CLI paramters and invoke `stateDB.RawDumpWith()` method.

PR #408, opened 2025-07-04
Feat/amain 410 control compaction

hayarobi:
This PR apply aergo-lib-1.3.0 to improve controlling BadgerDB.


kroggen:
My question is this:

With these changes, will `aergosvr` continue working with other db engines?

If I swap BadgerDB for LevelDB, will it work normally as expected?

If there is functionality that is specific to some db engine, it should be enclosed in conditional blocks

cmd/aergosvr/aergosvr.go
```diff
@@ -118,6 +120,14 @@ func rootRun(cmd *cobra.Command, args []string) {
...
 	pmapSvc := polarisclient.NewPolarisConnectSvc(cfg.P2P, p2pSvc)
 
+	chainSvc.SDB().MaintenanceEvent = func(event db.CompactionEvent) {
+		svrlog.Info().Msg("Maintenance event Started")
+		svrlog.Info().Msg("Closing listener")
+		rpcSvc.Listener.Close()
+		time.Sleep(10 * time.Second)
+		svrlog.Info().Msg("Connection drained for 10 seconds")
+	}
+
```
kroggen, line 130: What is the reason of this code block?
It disables the RPC when the db engine is doing background work?

hayarobi:
> My question is this:
> 
> With these changes, will `aergosvr` continue working with other db engines?
> 
> If I swap BadgerDB for LevelDB, will it work normally as expected?
> 
> If there is functionality that is specific to some db engine, it should be enclosed in conditional blocks

I guess it work.  
This PR mainly changes two things; 
One is changing max level to 8, and this affects the node itself only. This node can participate to chain with older versions of node.
Another is  Option feature of compaction control and monitoring, and this affects the badger implementation only. The other DB implementations will ignore this Option fields, so they woks samely as before.

cmd/aergosvr/aergosvr.go
hayarobi: Actually, most of works in this PR was done by @shepelt , and my works are the boilerplate and housekeeping to fit in v2.8.0 .

@shepelt I couldn't figure out all of changes in detail to explain, but I think you have reason to do that. Can you give a brief explanation?

chain/chainservice.go
```diff
@@ -69,9 +70,10 @@ func NewCore(dbType string, dataDir string, testModeOn bool, forceResetHeight ty
 }
 
 // Init prepares Core (chain & state DB).
-func (core *Core) init(dbType string, dataDir string, testModeOn bool, forceResetHeight types.BlockNo) error {
+func (core *Core) init(dbType string, dataDir string, testModeOn bool, forceResetHeight types.BlockNo, config *cfg.DBConfig) error {
 	// init chaindb
-	if err := core.cdb.Init(dbType, dataDir); err != nil {
+	if err := core.cdb.Init(dbType, dataDir, db.WithControlCompaction(
+		config.ControlCompaction, config.ChainDBPort)); err != nil {
```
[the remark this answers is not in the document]
hayarobi: I think that it is better the caller not to know detailed implementation of the callee, making loosely coupled between them. 
Though this feature only supports BadgerDB right now, but we may add this feature to other DBs if needed because the interface itself is defined in aergo-lib.
How about to append cooment to inform, like this; `// This option is only works only on BadgerDB in aergo 2.8`.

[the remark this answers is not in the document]
kroggen: OK, and separate the code into 2 lines
```go
// Compaction option currently only supported by BadgerDB
options := db.WithControlCompaction(config.ControlCompaction, config.ChainDBPort);
if err := core.cdb.Init(dbType, dataDir, options); err != nil {
```

kroggen (approved):

cmd/aergosvr/aergosvr.go
hayarobi: It was answered by other communication channels.

PR #423, opened 2025-07-26
Fix/unittest pr418

hayarobi:
This will fix compile error and failures while running unit test.


kroggen (approved):
I guess this can be merged because it does not affect the RPC
It worked with herajs

PR #428, opened 2025-09-25
chore: remove unused min/max func

cargoedit:
Go 1.21 adds built-in functions min and max.   https://tip.golang.org/doc/go1.21
Therefore, we can directly remove our own implementations and use the built-in function with the same names.


kroggen:
This is all you have to add? Seriously?
It is not even adding anything, but rather removing.
This can only be automated

cargoedit:
> This is all you have to add? Seriously? It is not even adding anything, but rather removing. This can only be automated

The Go standard library implements a method with the same name, so it can be removed. You can refer to the link above for more details.

PR #433, opened 2026-01-14
fix build on new go versions

kroggen:
The build is failing on go 1.24 with the error message:

```
contract/lstate_factory.go:79:10: cannot define new methods on non-local type LState
```

Go 1.24 made stricter enforcement around type aliases

The fix is very simple. This function appears to not even be used.


hayarobi (approved):
lgtm

PR #435, opened 2026-04-15
fix: raft snapshot sync

kroggen:
Other 2 branches on top of this one:

* [fix/raft-snapshot-sync-hashtabledb](https://github.com/aergoio/aergo/tree/fix/raft-snapshot-sync-hashtabledb)

* [fix/raft-consensus-2.8](https://github.com/aergoio/aergo/tree/fix/raft-consensus-2.8)

### DIFF
--- a/consensus/impl/raftv2/blockfactory.go
+++ b/consensus/impl/raftv2/blockfactory.go
@@ -757,37 +757,76 @@ func (bf *BlockFactory) MakeConfChangeProposal(req *types.MembershipChange) (*co
 }
 
 // getHardStateOfBlock returns (term/commit) corresponding to best block hash.
-// To get hardstateinfo, it needs to search all raft indexes.
+//
+// Lookup strategy, from most to least precise:
+//  1. Exact match: the raft entry for the requested block hash is still in WAL.
+//  2. Upward scan: the requested block's entry has been compacted, but a newer
+//     block's entry is still in WAL. Return that — it represents the oldest
+//     raft index that the caller can safely resume from (any older index has
+//     been compacted here and cannot be served via MsgApp anyway).
+//  3. Snapshot fallback: no block entry in WAL at all (deeply lagging caller,
+//     or fresh cluster). Return this node's latest snapshot metadata, which is
+//     by definition a valid raft checkpoint.
+//
+// Prior to this, only (1) and a downward scan (which can never succeed if the
+// caller's block is pre-retention — every older block is also pre-retention)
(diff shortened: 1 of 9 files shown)


consensus/impl/raftv2/raftserver.go
```diff
@@ -1273,6 +1273,16 @@ func (rs *raftServer) Process(ctx context.Context, m raftpb.Message) error {
 	if node == nil {
 		return ErrRaftNotReady
 	}
+	// A MsgHeartbeat from the leader carries the leader's current commit index.
```
hayarobi, line 1276: I am concerned that this modification might cause other side effects. I think it's better waiting for the synchronization tests on some more cases.

consensus/impl/raftv2/blockfactory.go
```diff
@@ -757,37 +757,76 @@ func (bf *BlockFactory) MakeConfChangeProposal(req *types.MembershipChange) (*co
...
 
-	logger.Warn().Uint64("request no", bestBlock.BlockNo()).Msg("can't find raft entry for requested hash. so try to find closest raft entry.")
+		logger.Warn().Uint64("request no", bestBlock.BlockNo()).Msg("raft entry for requested block hash is not in WAL; scanning newer blocks")
 
-	// find best hash mapping (no < bestBlock no)
-	for i := bestBlock.BlockNo() - 1; i >= 1; i-- {
-		if hash, err = bf.GetHashByNo(i); err == nil {
-			if entry, err = bf.ChainWAL.GetRaftEntryOfBlock(hash); err == nil {
-				logger.Debug().Str("entry", entry.ToString()).Msg("find best closest entry")
+		// (2) Upward scan.
```
hayarobi: I suspect that this code has a bug. 
It read my bestBlockNumber, so that block is the highest block, but it then scan upward block. There is no upward block in this node, so that I think this block always result in fail.

kroggen: The `bestBlockHash` parameter is not from this node's own best block, it is from the requester peer (the recovering node) best block hash, arriving over p2p.

In the recovery scenario the point of the function is that this node (provider) is ahead of the requester node, so there are plenty of upward blocks on this node to feed to the peer.

kroggen: I renamed the variables to make this cleaner

hayarobi: You're right. I made a mistake while analyzing the code.

hayarobi (approved):
lgtm

PR #436, opened 2026-04-17
fix: verify txn from named account

kroggen:


chain/signVerifier.go
```diff
@@ -138,7 +138,7 @@ func (sv *SignVerifier) verifyTx(comm component.IComponentRequester, tx *types.T
 			logger.Error().Err(err).Msg("failed to get verify because of opening contract error")
 			return false, err
 		}
-		address := name.GetOwner(cs, tx.Body.Account)
+		address := name.GetAddress(cs, tx.Body.Account)
```
hayarobi, line 141: name.GetAddress returns address or special name -aergo.system, aergo.name or etc - as is; that means it doesn't always return wallet address, while GetOwner always returns address, AFAIK. The function VerifyTxWithAddress take the address and then try to derive pubkey from address. I think that the special name will cause failure. Could you test with this?

kroggen: This line is converting the name from the transaction signer account into the corresponding address

`aergo.system`, `aergo.name` or etc cannot sign transactions

hayarobi: Looking inside the VerifyTxWithAddress function, it returns an error when aergo.system is passed as a parameter. The operation is fine, but the flow feels somewhat less clean to me.

hayarobi (approved):
lgtm

chain/signVerifier.go
kroggen: If someone fills the transaction's sender field with either `aergo.system` or `aergo.name`, the transaction is expected to be rejected by the nodes. And it would happen in many different ways, because even the signature of the transaction cannot be valid. These "names" correspond to internal contracts, not EOAs

PR #438, opened 2026-04-24
fix: update function comment for NewNTContainer

purelualight:
update function comment for NewNTContainer


purelualight:
@hayarobi Hi, Could you please review this PR at your convenience? Thank you very much.

PR #440, opened 2026-07-20
fix raft snapshot sync - 2.3.5 - part 2

kroggen:


consensus/impl/raftv2/raftserver.go
```diff
@@ -1263,6 +1263,8 @@ func (rs *raftServer) publishEntries(ents []raftpb.Entry) bool {
 
 				if block != nil {
 					if isDuplicateCommit(block) {
+						// raft index must advance even when the block is skipped
```
hayarobi, line 1266: Is there any chance that Index of duplicated commit is lower than current `rs.appliedIndex`?

kroggen: No — when that duplicate path runs, `ents[i].Index` is always greater than the current `rs.appliedIndex`, never lower

kroggen: Every batch goes through entriesToApply first:

```go
func (rs *raftServer) entriesToApply(ents []raftpb.Entry) (nents []raftpb.Entry) {
      if len(ents) == 0 {
          return
      }
      firstIdx := ents[0].Index
      if firstIdx > rs.appliedIndex+1 {
          logger.Fatal().Msgf("first index of committed entry[%d] should <= progress.appliedIndex[%d] 1", firstIdx,
  rs.appliedIndex)
      }
      if rs.appliedIndex-firstIdx+1 < uint64(len(ents)) {
          nents = ents[rs.appliedIndex-firstIdx+1:]
      }
      return nents
  }
  ```

hayarobi (approved):
LGTM

PR #441, opened 2026-07-23
Raft Consensus Hardening - 2.3.5

kroggen:
# Raft Consensus Hardening

Reviewer notes for the security and correctness changes on top of `release/2.3.5`.

Branch: `fix/raft-consensus-2.3.5`

**Scope:** permissioned / private Raft deployments. These commits harden trust boundaries around Raft messaging, snapshots, join/recovery, and a few race/liveness bugs.

**Not included:** gas / economics changes

**Compatibility (short):** no block/tx format or state-transition hardfork. Trusted Raft members can roll binaries gradually; coordinate **PreVote** carefully (election behavior). See per-commit notes below.

---

## Commit map

| # | Commit | One-line |
|---|--------|----------|
| 1 | `b3881cd7` | bind raft messages to authenticated peer ids |
| 2 | `64d481c7` | reject unauthorized snapshots and limit snapshot allocations |
| 3 | `54cb50cc` | require quorum agreement on cluster bootstrap |
| 4 | `ac1ede38` | reject peer id mismatch when joining existing cluster |
| 5 | `e4ab2492` | sync backup chain forward before exact raft hardstate reset |
| 6 | `22bad49e` | require block signer to be a current raft member |
| 7 | `31d3bd30` | enable raft prevote |
| 8 | `a555c025` | notify syncer on busy paths and verify sync target hash |
| 9 | `6346582f` | fix leader status locking and commit progress races |
| 10 | `ee88e34b` | validate membership and conf-change inputs |

---

## 1. Bind Raft messages to authenticated peer IDs

**Commit:** `bind raft messages to authenticated peer ids`

### Problem

Raft envelopes (`raftpb.Message`) were accepted from any connected libp2p peer without checking that:

- the peer is a current cluster member, and
- `Message.From` / `Message.To` match the authenticated peer identity and this node.

A connected non-member (or a member spoofing another member’s Raft ID) could inject votes, heartbeats, or append messages and disrupt elections or log replication.

### Fix

- Added `AergoRaftAccessor.ValidateMessage(peerID, msg)`.
- `Process` validates before stepping the Raft state machine.
- Wrapper handler rejects payloads from non-members early (`ParsePayload`).
- Binding rules: peer must map to a member; `From` must equal that member’s Raft ID; `To` must equal the local Raft ID.

### Alternatives considered

| Alternative | Why not selected |
|-------------|------------------|
| Only check membership, ignore `From`/`To` | Still allows a member to spoof another member’s Raft identity |
| Mutual TLS / separate Raft port with static certs | Large operational change; libp2p already authenticates peer IDs |
| Gate behind a hardfork height | This is P2P admission, not block validity; delaying weakens security on upgrade |

### Review focus

- Honest member→member traffic still has correct `From`/`To` in the existing transport.
- Joining / recovering nodes that are not yet in membership cannot send Raft msgs until merged (intentional).

---

## 2. Reject unauthorized snapshots and limit snapshot allocations

**Commit:** `reject unauthorized snapshots and limit snapshot allocations`

### Problem

Several snapshot weaknesses:

1. Raft message envelope decode used an effectively unbounded size (`1<<63`), enabling allocation DoS.
2. Snapshot streams were not clearly restricted to cluster members; concurrent receives were unbounded.
3. Snapshot contents were not checked for membership/`ConfState` consistency before/during apply.
4. Chain sync for a snapshot did not prefer the snapshot sender; publish path did not verify the synced block hash against snapshot chain data.
5. Snapshot response body length was not capped on the sender side.

### Fix

- Cap Raft envelope decode (`maxSnapshotRaftMessageSize`) and snapshot response body size.
- Reject snapshot streams from non-members; allow only one concurrent snapshot receive.
- `ValidateMessage` for `MsgSnap` (stale term/index, non-leader sender when leader known, membership consistency).
- `validateSnapshotConsistency` + publish-time block hash check.
- Sync chain from the snapshot sender when possible; lock-safe `getMemberPeerAddress`.

### Alternatives considered

| Alternative | Why not selected |
|-------------|------------------|
| Keep unbounded decode “because DB snapshot is large” | Envelope is small; DB stream is separate — unbounded decode was unnecessary |
| Only rate-limit without auth | Does not stop malicious snapshot content from a non-member |
| Require hardfork to enable | Same as (1): defense should apply on binary upgrade |

### Review focus

- Confirm envelope limit does not break legitimate large *DB* snapshot transfers (stream after envelope).
- Concurrent snapshot limit may reject overlapping catch-ups under load (fail closed).

---

## 3. Require quorum agreement on cluster bootstrap

**Commit:** `require quorum agreement on cluster bootstrap`

### Problem

`ConcurrentClusterInfoReceiver` selected the response with the highest `BestBlockNo`. A single malicious or partitioned peer advertising an extreme height (and arbitrary membership / hardstate) could win bootstrap and feed a joining node a false cluster view or unsafe checkpoint.

### Fix

- Fingerprint responses (chain ID, cluster ID, canonical member set, hardstate term/commit).
- Require at least `len(peers)/2+1` agreeing responses.
- Among the agreeing set, pick the highest `BestBlockNo`.

### Alternatives considered

| Alternative | Why not selected |
|-------------|------------------|
| Require unanimous agreement | Fragile if one peer is down or slightly divergent |
| Trust only designated seed peers | Extra config; quorum of connected peers matches Raft threat model better |
| Keep highest tip but verify signatures on membership | No signed cluster attestation exists today |

### Review focus

- Join needs enough live peers for quorum; small/partitioned peer sets may fail closed (safer).
- Hardstate is part of the fingerprint: disagreeing checkpoints cannot form a majority.

---

## 4. Reject peer ID mismatch when joining existing cluster

**Commit:** `reject peer id mismatch when joining existing cluster`

### Problem

When merging into an existing cluster, a name match with a **different** libp2p peer ID was only logged. Join could continue with the wrong identity mapping, causing later auth/transport confusion or impersonation of a member slot.

### Fix

`ValidateAndMergeExistingCluster` returns `false` if the remote member’s peer ID does not match this node’s peer ID (fatal on import).

### Alternatives considered

| Alternative | Why not selected |
|-------------|------------------|
| Auto-rewrite remote peer ID to local | Silently changes cluster identity; dangerous |
| Allow mismatch with operator flag | Easy to misuse during DR; explicit config of the correct peer key is clearer |

### Review focus

- Disaster recovery that promotes a full-node backup onto a BP must use the **member slot’s** peer key/identity.

---

## 5. Sync backup chain forward before exact Raft hardstate reset

**Commit:** `sync backup chain forward before exact raft hardstate reset`

### Problem

`usebackup` recovery asked live BPs for a Raft hardstate matching the backup tip, then `ResetWAL`’d to that checkpoint.

Previously, if the exact entry was missing (WAL compact), providers could return a **newer** entry or the latest snapshot. That pairs a high Raft commit/applied index with an older chain tip and can let the recovering node **vote or lead before it has that state**.

A first hardening attempt that **only** accepted exact hardstate (fail otherwise) was too harsh for DR: full-node backups are often older than retained Raft history, and sync-from-genesis is too slow.

### Fix (selected)

Keep **exact-only** hardstate for `ResetWAL`, but make recovery fast:

1. Providers still return membership if exact hardstate is unavailable (non-fatal).
2. Recovering node with `usebackup` and no exact hardstate **syncs only the delta** backup tip → live cluster tip.
3. Retries hardstate for the new tip, then `ResetWAL`.

So Raft never jumps ahead of the chain; DR stays incremental-sync fast.

### Alternatives considered

| Alternative | Pros | Cons / why not primary |
|-------------|------|-------------------------|
| Exact hardstate or fail | Safest one-liner | Rejects typical stale full-node backups |
| Restore “newer hardstate” fallback | Fast one-shot | Unsafe vote/lead with fabricated applied state |
| Newer hardstate + stay non-promotable until catch-up | Fast + safer | More state machine complexity; Tick gating alone is easy to get wrong |
| Empty WAL + leader `MsgSnap` only | No invented hardstate | Depends entirely on snapshot path reliability |
| **Sync-forward then exact reset (chosen)** | Fast delta sync; preserves exactness | Needs live peers with tip status; brief wait if LastStatus not ready |

### Review focus

- Confirm `ClusterInfo` soft-fail of hardstate does not weaken join when membership itself is wrong (still needs quorum agreement from commit 3).
- Sync target is chosen from **cluster member** peers’ `LastStatus` (not arbitrary watchers).
- After sync, tip must still be within retained Raft history on providers (normally true for current tip).

---

## 6. Require block signer to be a current Raft member

**Commit:** `require block signer to be a current raft member`

### Problem

`IsBlockValid` only ensured the block public key parsed. Any key with a valid signature could pass consensus-level validation, even if the producer was not a current Raft member. That matters if a non-member block is injected via P2P `AddBlock` / sync paths that call `IsBlockValid` during execute.

### Fix

Reject blocks whose producer peer ID is not in the current cluster member set (`bf.bpc`).

### Alternatives considered

| Alternative | Why not selected / caveat |
|-------------|---------------------------|
| Validate membership **at block height** (historical) | More correct across add/remove; needs membership history — not available as a simple API today |
| Hardfork-gated activation | Cleaner on public networks; for trusted private Raft, tip blocks are already member-signed, so immediate enable is usually fine |
| Rely only on Raft log commit path | Misses non-Raft P2P block injection paths |

### Review focus

- **Caveat:** check uses *current* membership. Deep sync/re-execute of blocks signed by a since-removed member can fail. Worth a follow-up if you reorg/sync across membership changes often.
- On a healthy Raft tip (leader ∈ members), old and new binaries agree — no practical tip fork among trusted BPs.

---

## 7. Enable Raft PreVote

**Commit:** `enable raft prevote`

### Problem

Without PreVote, a partitioned or restarting member can start a real election, inflate its term, then rejoin and force the leader to step down even when it cannot win. That causes unnecessary leadership flaps (classic Raft disruption).

### Fix

Set `PreVote: true` in Raft config. Campaigners must win a pre-vote (without bumping term) before a real election.

### Alternatives considered

| Alternative | Why not selected |
|-------------|------------------|
| Leave PreVote off | Keeps known disruption mode |
| Cluster-version negotiated enable (etcd-style) | Aergo Raft has no cluster feature negotiation today |
| Hardfork height gate | PreVote is process-wide election behavior, not per-block validation |

### Review focus

- **Rolling upgrade:** mixed PreVote / non-PreVote members can behave oddly during elections. Prefer upgrading all Raft members in a short window.
- PreVote does **not** stop a malicious member that is caught up and can win a majority — that is inherent to Raft.
- Combined with commit 1, outsiders should not inject votes at all.

---

## 8. Notify syncer on busy paths and verify sync target hash

**Commit:** `notify syncer on busy paths and verify sync target hash`

### Problem

1. If the syncer was already running or the target was already reached, callers waiting on `NotifyC` could hang (no notification).
2. After sync, consensus did not verify that the block at `targetNo` matched the requested `targetHash` (snapshot / consensus sync could accept the wrong block at that height).

### Fix

- Notify waiters on busy / already-synced / error paths (`ErrSyncerBusy` when busy).
- `SyncChain` verifies hash after sync (and if already at/above target).
- Use buffered notify channel.

### Alternatives considered

| Alternative | Why not selected |
|-------------|------------------|
| Timeout-only on waiter side | Masks the bug; still races |
| Verify only inside syncer | Consensus callers still need a hash guarantee at their API |

### Review focus

- Backup recovery (commit 5) and snapshot sync both depend on reliable `SyncChain` completion notification.

---

## 9. Fix leader status locking and commit progress races

**Commit:** `fix leader status locking and commit progress races`

### Problem

Several concurrency bugs:

- `updateLeader` / `GetLeaderStatus` used the wrong mutex / returned a copy that included the mutex unsafely.
- `CommitProgress` getters returned pointers to internal fields under brief lock (TOCTOU / race with `-race`).
- Snapshot sender could leave the write goroutine orphaned on stop/timeout (race / leak).
- `IsIDRemoved` / member lookups lacked consistent locking in some paths.
- Minor nil guard on duplicate-commit detection when request progress is empty.

### Fix

- Lock `leaderStatus` properly; return a value copy of status fields.
- Return copies from `GetConnect` / `GetRequest`.
- Join writer completion on snapshot send cancel/timeout; cap already added in commit 2.
- Lock cluster for remove-ID / member getters as needed.

### Alternatives considered

| Alternative | Why not selected |
|-------------|------------------|
| Channel-serialize all Raft status access | Larger refactor |
| Ignore races under “Raft is single-threaded” | Status and P2P run concurrently with the ready loop |

### Review focus

- Run with `-race` on `raftv2` / `raftsupport` if not already in CI for this branch.

---

## 10. Validate membership and conf-change inputs

**Commit:** `validate membership and conf-change inputs`

### Problem

Membership and conf-change paths trusted partially validated attributes:

- Peer IDs were not parsed in `Member.IsValid`.
- Nil `MembershipChange` / attrs could panic or proceed oddly.
- Snapshot recover / add paths did not consistently validate members.
- Malformed conf-change protobuf context could `Fatal` the process instead of returning an error.

### Fix

- Validate peer ID bytes in `Member.IsValid`.
- Nil-guard membership RPCs / `makeProposal` / `NewMemberFrom*`.
- Validate members when recovering from snapshot.
- Unmarshal conf-change errors become returned errors (no process fatal on bad entry).

### Alternatives considered

| Alternative | Why not selected |
|-------------|------------------|
| Keep Fatal on bad conf-change | Turns corrupt/malicious log data into permanent outage |
| Strict validate only on AddMember RPC | Leaves snapshot recover and WAL replay gaps |

### Review focus

- Existing clusters with historically weird peer ID encodings would fail `IsValid` — unlikely if peers were always libp2p IDs.

---

## Suggested test plan for reviewers

- [ ] Unit: `go test ./consensus/impl/raftv2/ ./p2p/raftsupport/ ./p2p/subproto/ ./consensus/chain/ ./syncer/`
- [ ] Race: same packages with `-race` (especially snapshot sender + leader status).
- [ ] 3-node Raft smoke: rolling restart of one BP; confirm election stability (PreVote).
- [ ] Join new member with correct peer ID; confirm mismatch is rejected.
- [ ] `usebackup` DR: copy slightly lagging full-node data to a BP slot; confirm delta sync then WAL reset (not fail-closed on “old tip”).
- [ ] Negative: non-member peer cannot drive Raft votes/snapshots.
- [ ] Snapshot catch-up of a lagging follower still succeeds under size limits.

---

## Residual risks / follow-ups

1. **`IsBlockValid` uses current membership** — consider height-indexed membership if deep sync across removes is required.
2. **PreVote rolling upgrade** — document ops procedure (brief coordinated restart).
3. **No cluster feature negotiation** — unlike etcd, flags flip per binary.
4. **Trusted-member model** — a compromised BP that holds a majority vote can still elect itself; these patches remove *outsider* and *spoofing* paths, not Byzantine majority failure.

### DIFF
--- a/consensus/chain/block.go
+++ b/consensus/chain/block.go
@@ -1,6 +1,7 @@
 package chain
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"time"
@@ -209,7 +210,24 @@ func SyncChain(hs *component.ComponentHub, targetHash []byte, targetNo types.Blo
 	logger.Info().Str("peer", p2putil.ShortForm(peerID)).Uint64("no", targetNo).
 		Str("hash", enc.ToString(targetHash)).Msg("request to sync for consensus")
 
-	notiC := make(chan error)
+	verifyTarget := func() error {
+		result, err := hs.RequestFuture(message.ChainSvc, &message.GetBlockByNo{BlockNo: targetNo},
+			time.Second, "consensus/chain.SyncChain.verifyTarget").Result()
+		block, err := message.GetHelper().ExtractBlockFromResponseAndError(result, err)
+		if err != nil {
+			return err
+		}
+		if block == nil || !bytes.Equal(block.BlockHash(), targetHash) {
+			return fmt.Errorf("%w: block %d hash does not match snapshot", ErrSyncChain, targetNo)
+		}
+		return nil
+	}
+
+	if best := GetBestBlock(hs); best != nil && best.BlockNo() >= targetNo {
+		return verifyTarget()
+	}
+
+	notiC := make(chan error, 1)
 	hs.Tell(message.SyncerSvc, &message.SyncStart{PeerID: peerID, TargetNo: targetNo, NotifyC: notiC})
 
 	// wait end of sync every 1sec
@@ -224,7 +242,13 @@ func SyncChain(hs *component.ComponentHub, targetHash []byte, targetNo types.Blo
 		}
 	}
 
+	if err := verifyTarget(); err != nil {
+		logger.Error().Err(err).Uint64("no", targetNo).
+			Str("hash", enc.ToString(targetHash)).
+			Msg("sync completed at a different block")
+		return err
+	}
+
 	logger.Info().Str("peer", p2putil.ShortForm(peerID)).Msg("succeeded to sync for consensus")
-	// TODO check best block is equal to target Hash/no
 	return nil
 }
--- a/consensus/chain/block.go
+++ b/consensus/chain/block.go
@@ -223,23 +223,32 @@ func SyncChain(hs *component.ComponentHub, targetHash []byte, targetNo types.Blo
 		return nil
 	}
 
-	if best := GetBestBlock(hs); best != nil && best.BlockNo() >= targetNo {
-		return verifyTarget()
-	}
-
-	notiC := make(chan error, 1)
-	hs.Tell(message.SyncerSvc, &message.SyncStart{PeerID: peerID, TargetNo: targetNo, NotifyC: notiC})
-
-	// wait end of sync every 1sec
-	select {
-	case err := <-notiC:
-		if err != nil {
-			logger.Error().Err(err).Uint64("no", targetNo).
-				Str("hash", enc.ToString(targetHash)).
-				Msg("failed to sync")
-
-			return err
+	for {
+		if best := GetBestBlock(hs); best != nil && best.BlockNo() >= targetNo {
+			return verifyTarget()
 		}
+
+		notiC := make(chan error, 1)
+		hs.Tell(message.SyncerSvc, &message.SyncStart{PeerID: peerID, TargetNo: targetNo, NotifyC: notiC})
+
+		err := <-notiC
+		if err == nil {
+			break
+		}
+		// Another sync is already running (e.g. orphan/tip-triggered). Wait
+		// for progress and retry instead of failing consensus sync.
+		if errors.Is(err, message.ErrSyncerBusy) {
+			logger.Info().Uint64("no", targetNo).
+				Str("hash", enc.ToString(targetHash)).
+				Msg("syncer busy; waiting to retry consensus sync")
+			time.Sleep(time.Second)
+			continue
+		}
+
+		logger.Error().Err(err).Uint64("no", targetNo).
+			Str("hash", enc.ToString(targetHash)).
+			Msg("failed to sync")
+		return err
 	}
 
 	if err := verifyTarget(); err != nil {
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -97,6 +97,7 @@ type ChainDB interface {
 
 // AergoRaftAccessor is interface to access raft messaging. It is wrapping raft message with aergo internal types
 type AergoRaftAccessor interface {
+	ValidateMessage(peerID types.PeerID, m raftpb.Message) error
 	Process(ctx context.Context, peerID types.PeerID, m raftpb.Message) error
 	IsIDRemoved(peerID types.PeerID) bool
 	ReportUnreachable(peerID types.PeerID)
--- a/consensus/impl/raftv2/blockfactory.go
+++ b/consensus/impl/raftv2/blockfactory.go
@@ -692,6 +692,9 @@ func (bf *BlockFactory) ConfChange(req *types.MembershipChange) (*consensus.Memb
 	if bf.bpc == nil {
 		return nil, ErrorMembershipChange{ErrClusterNotReady}
 	}
+	if req == nil || req.Attr == nil {
+		return nil, ErrorMembershipChange{consensus.ErrInvalidMemberAttr}
+	}
 
 	if !bf.raftServer.IsLeader() {
 		return nil, ErrorMembershipChange{ErrNotRaftLeader}
@@ -727,6 +730,9 @@ func (bf *BlockFactory) MakeConfChangeProposal(req *types.MembershipChange) (*co
 
 	if bf.bpc == nil {
 		return nil, ErrorMembershipChange{ErrClusterNotReady}
+	}
+	if req == nil || req.Attr == nil {
+		return nil, consensus.ErrInvalidMemberAttr
 	}
 
 	cl := bf.bpc
--- a/consensus/impl/raftv2/blockfactory.go
+++ b/consensus/impl/raftv2/blockfactory.go
@@ -34,10 +34,11 @@ var (
 )
 
 var (
-	ErrClusterNotReady      = errors.New("cluster is not ready")
-	ErrNotRaftLeader        = errors.New("this node is not leader")
-	ErrInvalidConsensusName = errors.New("invalid consensus name")
-	ErrCancelGenerate       = errors.New("cancel generating block because work becomes stale")
+	ErrClusterNotReady         = errors.New("cluster is not ready")
+	ErrNotRaftLeader           = errors.New("this node is not leader")
+	ErrInvalidConsensusName    = errors.New("invalid consensus name")
+	ErrCancelGenerate          = errors.New("cancel generating block because work becomes stale")
+	ErrUnauthorizedBlockSigner = errors.New("block signer is not a current raft member")
 )
 
 func init() {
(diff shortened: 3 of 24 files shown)


hayarobi:
There is an issue about dependent version of aergo-lib. Maybe tag or commit of aergo-lib is not synced yet. 

Please ses https://github.com/aergoio/aergo/actions/runs/31240027540

kroggen:
The Actions is for another PR, not this one

I will try to fix it anyway

kroggen:
it was lacking to run `go mod tidy` on both branches

done

kroggen:
Now it is failing with "database or disk is full" on `TestTypeBigTable`:

```
--- FAIL: TestTypeBigTable (11.02s)
    vm_dummy_pub_test.go:777: 
        	Error Trace:	/home/runner/work/aergo/aergo/contract/vm_dummy/vm_dummy_pub_test.go:777
        	Error:      	Received unexpected error:
        	            	AmgGDzeTLZu2Sm4KF42a1GR4hyMTymQgv8DinRx9JmPhfeeTv9ii:0: database or disk is full
        	Test:       	TestTypeBigTable
        	Messages:   	failed to call tx
```

But it does not fail locally

Maybe GitHub Actions machine is resources restrained?

hayarobi (approved):
Good job!

PR #442, opened 2026-07-23
Raft Consensus Hardening - 2.8

kroggen:
Includes HashTableDB

Similar to the PR #441 that applies on v2.3.5

This one runs the integration tests


hayarobi (approved):
I think that this PR itself is enough to approve, but it is needed to investigate the failure of the complete test.

hayarobi (approved):
LGTM

PR #444, opened 2026-08-18
Make legacy branch testable

hayarobi:
It appears that actions on branches derived from the release/2.3.x branch will execute as expected only if this PR is merged, since any of GitHub actions in all branches are executed based on files in the base branch,

This PR itself only changes a single environment variable, so it is no harm to other branches, I guess.

### DIFF
--- a/.github/workflows/full_test.yml
+++ b/.github/workflows/full_test.yml
@@ -18,6 +18,9 @@ on:
 jobs:
   build_and_test:
     runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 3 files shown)


kroggen:
this change does not need to be made directly on the tested branch?

this PR will merge into `develop`

hayarobi:
My bad, I misunderstood the docs. I had assumed the workflow would always run based on the file in the default branch, no matter which branch triggered it.

PR #446, opened 2026-09-03
fix small memory leak on contract deploy

### LINKED ISSUE #445: Possible memory leak of `addr` on the constructor-error paths of `luaDeployContract`
# Possible memory leak of `addr` on the constructor-error paths of `luaDeployContract`

`luaDeployContract` allocates the new contract address with `C.CString` and returns it to
the C caller, which frees it. Three early returns after that allocation return a different
string instead, and `addr` is never freed on those paths.

contract/vm_callback.go:1385

```go
	addr := C.CString(types.EncodeAddress(newContract.ID()))
	ret = C.int(1)

	if ce != nil {
		// run the constructor
		defer setInstCount(ce.ctx, L, ce.L)
		ret += ce.call(minusCallCount(ctx, C.vm_instcount(L), luaCallCountDeduc), L)

		// check if the execution was successful
		if ce.err != nil {
			// rollback the recovery point
			err := clearRecoveryPoint(L, ctx, seq, true)
			if err != nil {
				return -1, C.CString("[Contract.LuaDeployContract] recovery error: " + err.Error())
			}
			// log some info
			if ctx.traceFile != nil {
				_, _ = ctx.traceFile.WriteString(fmt.Sprintf("recovery snapshot: %d\n", seq))
			}
			// return the error message
			return -1, C.CString("[Contract.LuaDeployContract] call err:" + ce.err.Error())
		}
	}

	if seq == 1 {
		err := clearRecoveryPoint(L, ctx, seq, false)
		if err != nil {
			return -1, C.CString("[Contract.LuaDeployContract] recovery error: " + err.Error())
		}
	}

	return ret, addr
}
```

Only the success return at the bottom hands `addr` back. The returns at lines 1398, 1405
and 1412 replace it with a freshly allocated error message, and there is no `defer C.free`
covering `addr`.

Ownership of the returned string is the C caller's, and it does release it.
contract/contract_module.c:375:

```c
	ret = luaDeployContract(L, service, contract, json_args, amount);
	if (ret.r0 < 0) {
		free(json_args);
		reset_amount_info(L);
		strPushAndRelease(L, ret.r1);
		luaL_throwerror(L);
	}
```

`strPushAndRelease` frees whatever `r1` is, so the error message is accounted for. The
address string allocated earlier is not part of `r1` on those paths, so nothing frees it.

The line 1405 path is the reachable one: it is taken whenever a deployed contract's
constructor returns an error. `moduleDeploy` is the `contract.deploy` builtin, so any
contract that deploys another contract with a failing constructor leaks
`len(EncodeAddress(id))+1` bytes per call, and the call can be repeated.

Fix:

```go
	addr := C.CString(types.EncodeAddress(newContract.ID()))
	returnAddr := false
	defer func() {
		if !returnAddr {
			C.free(unsafe.Pointer(addr))
		}
	}()
```

with `returnAddr = true` immediately before `return ret, addr`, or by freeing `addr`
explicitly at each of the three early returns.

If you could credit me as a reporter for my contributions to security advisory I will be thankful.

kroggen:
Closes #445

### DIFF
--- a/contract/vm_callback.go
+++ b/contract/vm_callback.go
@@ -1382,7 +1382,6 @@ func luaDeployContract(
 	// increment the nonce of the creator
 	senderState.SetNonce(senderState.Nonce() + 1)
 
-	addr := C.CString(types.EncodeAddress(newContract.ID()))
 	ret = C.int(1)
 
 	if ce != nil {
@@ -1413,7 +1412,7 @@ func luaDeployContract(
 		}
 	}
 
-	return ret, addr
+	return ret, C.CString(types.EncodeAddress(newContract.ID()))
 }
 
 //export isPublic


hayarobi (approved):
Good

